PR #23, opened 2024-06-25
Fix #22: Add config for site and data directories

### LINKED ISSUE #22: Allow to use any directory as site root
This would be part of the Statiq config (Deployment)
Default would be `src/main/site` and `data` would be relative to it

This might change the way we scan for files, see Quinoa for example:
https://github.com/quarkiverse/quarkus-quinoa/blob/main/deployment/src/main/java/io/quarkiverse/quinoa/deployment/QuinoaProcessor.java#L358-L378

@melloware this one is for you I guess?

melloware:
Fix #22: Add config for site and data directories


melloware:
@ia3andy i started this to make sure it was what you were thinking.

melloware:
Refactored to `common-deployment` let me know how you want me to handle the new `data` Config that was added to roq-data?

ia3andy (approved):
Thanks!

I will use a BuildItem to keep the directories

PR #26, opened 2024-06-27
Refactor

ia3andy:


melloware (approved):
LGTM.  I wonder if we should have done something similar in Quinoa with a BuildItem then?

ia3andy:
@melloware now we should use this BuildItem in the roq-data

PR #104, opened 2024-09-23
Fix SEO links

ia3andy:
@melloware this is mega-awesome :)

I just plugged the collections pages prev and next.
I also added a small fix on the title (this was because my previous version that you used was not totally right)

### DIFF
--- a/roq-frontmatter/runtime/src/main/resources/templates/tags/seoTitle.html
+++ b/roq-frontmatter/runtime/src/main/resources/templates/tags/seoTitle.html
@@ -1,9 +1,12 @@
 {#if pageTitle || siteTitle}
   <!-- SEO TITLE -->
+  {#fragment title rendered=false}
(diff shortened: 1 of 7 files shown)


roq-frontmatter/runtime/src/main/resources/templates/tags/seoTitle.html
```diff
@@ -1,9 +1,12 @@
 {#if pageTitle || siteTitle}
   <!-- SEO TITLE -->
+  {#fragment title rendered=false}
+  {#if pageTitle && pageTitle ne siteTitle}{pageTitle} - {siteTitle}{#else}{siteTitle}{/if}
+  {/fragment}
   <title>
-    {#if pageTitle ne siteTitle}{pageTitle} - {siteTitle}{#else}{siteTitle}{/if}
+    {#include $title /}
   </title>
-  <meta name="twitter:title" content="{#if pageTitle ne siteTitle}{pageTitle} - {siteTitle}{#else}{siteTitle}{/if}">
```
ia3andy, line 6: @mkouba capturing in action :)

PR #117, opened 2024-09-26
Change the way we deal with data

ia3andy:
Due to an issue with Qute, I had to change the way we deal with data.

From now on, data is available under `site.data` or `page.data`.

Some shortcut methods allow to access specific data from site or page:
- url (the page url as a `RoqUrl`)
- img (the image url as a `RoqUrl`)
- title
- description

for NormalPage, we can access paginator (if defined)
for DocumentPage, we can also access the collection and the date

This PR also change a bit the way we deal with url making it simpler and more consistant, allowing to call absolute or relative at the end.


melloware (approved):
I actually like this better

ia3andy:
Yeah me too, it's more type-safe and consistant. I was stuck on keeping it similar to Jekyll

PR #119, opened 2024-09-27
Start doc

ia3andy:
this also contains using `@TemplateData` for frontmatter model


roq-generator/runtime/src/main/java/io/quarkiverse/roq/generator/runtime/RoqGeneratorConfig.java
```diff
@@ -14,25 +15,33 @@ public interface RoqGeneratorConfig {
...
-     *                 SelectedPath.builder().path("/roq?name=foo").build(),
-     *                 SelectedPath.builder().path("/roq?name=bar").build()));
-     *     }
-     * </code>
+     *
+     * <pre>
+     * {@code
+     * &#64;Produces
+     * &#64;Singleton
+     * @Transactional
```
melloware: I think `@Transactional` needs to be `&#64;Transactional`

melloware (changes requested):

roq-generator/runtime/src/main/java/io/quarkiverse/roq/generator/runtime/RoqGeneratorConfig.java
ia3andy: yeah it's the other way around, thanks, it's fixed

ia3andy: Ah no you were right, but weirdly process-sources changed them to `&#64;` but not @Transactional 🤨

PR #146, opened 2024-10-08
Add about page and more

### LINKED ISSUE #125: Create Credits page with contributors
Add all-contribs
Automate copy of all-contribs json files and print the content in the blog

ia3andy:
Closes #125 

Added:
- about page based on all-contributors levering awesome roq-data
- use roq-data for menu (there is a bug for a raw data array, when iterating, it doesn't wrap items content as JsonObject, it works with an extra `items` layer).
- favicon

image


melloware (approved):
That is so slick

PR #171, opened 2024-10-10
Fix #116: RSS Feed updates and blog post

### LINKED ISSUE #116: RSS Feed for Blog
One of the things Gatsby does is it generates this in the `<head>`

```xml
<link rel="alternate" type="application/rss+xml" title="Melloware Blog" href="[/rss.xml]"/>
```

and example of the RSS feed it generates is

https://www.melloware.com/rss.xml

melloware:
Fix #116: RSS Feed updates and blog post


ia3andy (approved):
Thanks

PR #176, opened 2024-10-14
Improvements

### LINKED ISSUE #172: Add `site.page(path)` and `site.document(path)`
This will allow to find pages by id. This way user won't have to resolve any url by themselves
For example:
```html
{site.page('rss.xml').url}
{site.page('/pages/about.html').url}

{site.document('_posts/my-post.md').url}
```

We should also remove the rootUrl from `site`, it is available from the `site.url.root` if really needed.

ia3andy:
Fixes #172
- Remove rootUrl from `site`, it is now hidden behind `site.url.root`, it is only necessary to access Quarkus resources.
- Add a way to access pages (and their url) from `site.page(id)` and `site.document(id)`
- Documentation on class and doc
- Change CollectionConfig to allow having hidden collection for data consumption only (like Jekyll does)


melloware (approved):
Good improvment!

PR #207, opened 2024-10-23
AsciiDoc native testing

melloware:
Just testing here to see if my fixes in Qute will work or if I need to add anything else


melloware:
OK good this fixes it so closing

melloware:
@ia3andy i added this here in case we don;t get it merged in Qute.

ia3andy:
Let's merge it and revert when we have it in Qute

PR #293, opened 2024-11-21
fix rendering for asciidoc templates

### LINKED ISSUE #290: QRCode plugin doesn't work with asciidoc plugin
It does with markdown which allows to write plain HTML but doesn't with asciidoc.

For asciidoc templates it should generate the image in the static folder and generate a link to it.

There is no access to the parent block in the section helper, so it's not possible to know what form should take the generated result.

We could add a type to the parameters to know if we should generate html or asciidoc code...

jtama:
Fix #290

### DIFF
--- a/docs/modules/ROOT/pages/quarkus-roq-plugins.adoc
+++ b/docs/modules/ROOT/pages/quarkus-roq-plugins.adoc
@@ -148,7 +148,11 @@ To install it add:
 
 Then create a template and add the `{#qrcode ...}` tag to it and style and size it as you want.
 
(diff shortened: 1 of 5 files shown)


melloware (changes requested):
Can you updates the QRCode section of the docs to mention this new asciidoc property and how to use it?

jtama:
> Can you updates the QRCode section of the docs to mention this new asciidoc property and how to use it?

Done, but not sure if the is english, or french translated to english ;) !

melloware (approved):

PR #382, opened 2025-01-28
Allow special characters, @RoqnRoll test and more

ia3andy:
- **Special Character Support:** Static file names now support special characters like accents and spaces (within valid file name constraints). This feature requires Quarkus 3.18+ due to issues in previous versions.  
- **New Testing Annotation:** Added `@RoqAndRoll` to test the full generation of Roq websites.  
- **Performance Improvement:** Static files are now copied directly instead of being served through HTTP, resulting in significant optimizations.  
- **Slug Update:** Slugs now allow the `_` character for improved flexibility.


ia3andy:
Build is finally fixed 🤯

PR #392, opened 2025-01-30
Rename `roq-docs` back to `docs`

gastaldi:
Otherwise build fails in https://github.com/quarkiverse/quarkiverse-docs

See https://github.com/quarkiverse/quarkiverse-docs/actions/runs/13052259452/job/36414995739


ia3andy:
@gastaldi let's change in the quarkiverse-docs to roq-docs

ia3andy:
that worked right? https://github.com/quarkiverse/quarkiverse-docs/commit/f7345fe4a2edfdb1670f12cc1a8f37fc637c6745

gastaldi:
@ia3andy yes, I did it so the build could work again. I'd recommend using `docs/`, as other extensions also use it and it's easier when we need to apply a patch

ia3andy:
> @ia3andy yes, I did it so the build could work again. I'd recommend using `docs/`, as other extensions also use it and it's easier when we need to apply a patch globally

ok :)

PR #399, opened 2025-02-05
Allow overriding "Contact me" label

jotak:
E.g. to replace with "Contact us"

Would `site.contact-title` work for you?

cc @ia3andy @melloware


ia3andy:
I think we should use i18n messages for this, let me check that works also for overriding defaults..

ia3andy:
@jotak we need this to allow using messages: https://github.com/quarkusio/quarkus/issues/46259.

I am sorry I will close until then to not introduce something that will be changed.

PR #411, opened 2025-02-19
Code snippets: dark theme and no word break

### LINKED ISSUE #404: Improve code snippets css (word-breaks etc.)
Code snippets are likely to have long lines, which aren't very well represented imho.

Currently we have:
Image

See how word-break happen even within a word. I think it should at least be `word-break: break-word;` to avoid that.

But I would go further and not break the lines, having scrollbars instead:

Image

Now the question is, what should be the default to have in roq? (I can do these changes in my blog css only, but I think roq defaults should change anyway)

jotak:
Fixes #404

It's mostly to fix the word-break problem mentioned in the issue.
With also a proposal to use a dark theme for hljs - that's a matter of preference of course, idk what you folks prefer

**Before:**
image

**After:**
image

### DIFF
--- a/roq-theme/default/src/main/web/roq.scss
+++ b/roq-theme/default/src/main/web/roq.scss
@@ -68,15 +68,9 @@ pre {
     padding-right: 0;
     padding-left: 0;
     overflow: auto;
-    white-space: pre-wrap;
-    word-wrap: break-word;
-    word-break: break-all;
-    background-color: #d8d8d8;
+    white-space: pre;
+    background-color: #222;
     border-radius: 5px;
-  }
-
-  code.hljs {
-    background-color: #d8d8d8;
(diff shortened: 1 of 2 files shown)


roq-theme/default/src/main/web/roq.scss
```diff
@@ -150,6 +144,7 @@ pre {
 }
 
 .post, .article-page, .page-content  {
+	width: 100%;
```
melloware: Looks like one minor spacing issue here but other than that looks good!

melloware (approved):

roq-theme/default/src/main/web/roq.scss
jotak: fixed

Issue #141, opened 2024-10-05
Generate an index as JSON for searching purposes


mcruzdev:
The idea is to generate an index from all generated HTML files to be used later by toolings like https://lunrjs.com/

ia3andy:
Go for it :)

ia3andy:
@mcruzdev any update on this?

ia3andy:
@mcruzdev ping

ia3andy:
FYI the sitemap should help you get started:
https://github.com/quarkiverse/quarkus-roq/blob/main/plugin/sitemap/runtime/src/main/resources/templates/fm/sitemap.xml

ia3andy:
@mcruzdev, @jtama is giving this a shot :)

PR #425, opened 2025-02-28
Adds lunr plugin to provide search capabilities

jtama:
The result is ugly as hell.

But it works.

Kind of.

Links #141 

@ia3andy


ia3andy:
😂 good start!

jtama:
clearly need help on the following:

- the js file is misplaced, but not sure if I can it to the runtime module `src\main\web`
- we should enhanced why a result has been proposed. There is a good inspiration => https://gitlab.com/antora/antora-lunr-extension/-/tree/main/data/js?ref_type=heads

ia3andy:
> Clearly need help on the following:
> 
> * the js file is misplaced, but not sure if I can it to the runtime module `src\main\web`
Yeah, I think we should be able to contribute to the user bundle, I'll investigate. This force the user to use the web-bundler, but I think it's fine.


> * we should enhanced why a result has been proposed. There is a good inspiration => https://gitlab.com/antora/antora-lunr-extension/-/tree/main/data/js?ref_type=heads
👍

> * Everything related to UI 😆

I will help you on that side, just make it work for now :)

jtama:
ok, it works.

It's still ugly, but it works, a great part of the js code is copy/pasted from the antora lunr extension, and it needs polish on the generated feed.

I didn't found a way to put the js where it belongs neither.

ia3andy:
It's getting pretty cool 😎

ia3andy (approved):
Thanks @melloware

PR #434, opened 2025-03-12
Update action.yml

DevrajKalpure:


action.yml
```diff
@@ -84,18 +86,17 @@ runs:
           exit 1
         fi
 
-    # Step 3: Setting environments for the build
```
ia3andy, line 87: Any reason to delete those steps comments?

ia3andy:
This is not the right file for this.

PR #467, opened 2025-04-07
Switch to Quarkus JGit extension

melloware:
This should do thing the Quarkus way and also provide GraalVM support


melloware:
cc @jtama

ia3andy:
@melloware we don't need the extension, we just use it at build item..

PR #493, opened 2025-04-28
Fix string splitting

### LINKED ISSUE #488: RoqTemplateExtension.asStrings() doesn't seem to support space-delimited strings
In a site I'm working on, we have tags defined in FM like this: 

```
tags: oidc Keycloak JWT encryption authentication
```

When the tagging plugin parses that via `RoqTemplateExtension.asStrings()` , we get a single tag back: `oidc-Keycloak-JWT-encryption-authentication`, which is obviously not correct. :) If I call `String.split(" ")` on the string, we get the expected array of 5 tags. I've looked at the regex used in the method, and it _seems_ like it should work, but it isn't for some reason, and regex fu is apparently not quite good enough to figure out why. :)

jasondlee:
Modify string split regex to cover more cases
Trim each tag
Filter out empty tags

Fixes #488

### DIFF
--- a/roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/RoqTemplateExtension.java
+++ b/roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/RoqTemplateExtension.java
@@ -1,5 +1,6 @@
 package io.quarkiverse.roq.frontmatter.runtime;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/RoqTemplateExtension.java
```diff
@@ -46,14 +47,25 @@ public static String slugify(String text) {
...
+                    .filter(s -> !s.isEmpty())
+                    .toList();
         }
         if (o instanceof JsonArray i) {
             return i.getList();
         }
         return List.of();
     }
 
+    public static void main(String[] args) {
```
ia3andy: Could we have a test instead :)

ia3andy (changes requested):

roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/RoqTemplateExtension.java
jasondlee: haha. I thought I deleted that. But yes.

jasondlee: Took me a minute to figure out the import sort plugin, but I think this oughtta do it, assuming the test is in the right place. :)

ia3andy (approved):
LGTM

PR #504, opened 2025-05-05
Add youtube icon on social section

mcruzdev:


ia3andy (commented):
Could we create a list of brands and then just iterate and check if it's in the data?

mcruzdev:
It is better.

mcruzdev:
> Could we create a list of brands and then just iterate and check if it's in the data?

It is very simple, but yes, can improve our exp a lot.

PR #507, opened 2025-05-07
Case-preserving slug

### LINKED ISSUE #506: Allow for mixed-case slugs
I have run into a scenario where I need a mixed case slug. For example, from the wildfly.org, if I have this in the FM:

```
link: /news/:year/:month/:day/:slug
```

A sample generated link looks this: http://localhost:8080/news/2025/05/05/exposing-wasm-binaries-as-mcp-tools/

However, in the live site, the link is this: https://www.wildfly.org/news/2025/05/05/Exposing-WASM-binaries-as-MCP-tools/

I have verified that case _does_ matter by trying this: https://www.wildfly.org/news/2025/05/05/exposing-wasm-binaries-as-mcp-tools/

In order to not break years of links, I'd like to be able to configure the slug method (https://github.com/quarkiverse/quarkus-roq/blob/main/roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/TemplateLink.java#L72-L85) to _not_ lowercase the generated slug. As a curious, if possibly irrelevant comparison, it seems `PathUtils.slugify()` (https://github.com/quarkiverse/quarkus-roq/blob/main/roq-common/runtime/src/main/java/io/quarkiverse/roq/util/PathUtils.java#L76-L84) doesn't lowercase at all. Seems odd, but there might be a reason I'm missing. Just noting this in case it matters. :)

Another possible alternative would be a placeholder for a case-preserving slug (e.g., `casedSlug`, `slug!`). 

I'm happy to send up a PR once I know what is acceptable.

jasondlee:
Add 'Slug' option for a case-preserving slug
Move '.toLowerCase()' outside of 'resolveSlug'
Update docs
Add test

Resolves #506

### DIFF
--- a/roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/TemplateLink.java
+++ b/roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/TemplateLink.java
@@ -62,7 +62,8 @@ public class TemplateLink {
                         () -> data.pageInfo().isHtml() ? ""
                                 : "." + data.pageInfo().sourceFileExtension()),
                 Map.entry(":ext!", () -> data.pageInfo().isHtml() ? ".html" : "." + data.pageInfo().sourceFileExtension()),
-                Map.entry(":slug", () -> resolveSlug(data))));
+                Map.entry(":slug", () -> resolveSlug(data).toLowerCase()),
+                Map.entry(":Slug", () -> resolveSlug(data)))); // Case-preserving slug
         if (other != null) {
             result.putAll(other);
         }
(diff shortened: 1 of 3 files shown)


roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/TemplateLink.java
```diff
@@ -81,7 +82,7 @@ public static String resolveSlug(LinkData data) {
                 title = baseFileName;
             }
         }
-        return slugify(title).toLowerCase();
```
jasondlee, line 84: I'm a _little_ concerned by this change, as the behavior of the method is changed, but it doesn't look like anything else in Roq itself is calling this, so it may be fine. Breaking this apart to preserve the behavior while implementing the feature seemed a little overkill, but I'm happy to do that if you feel we need to preserve behavior here.

ia3andy (approved):
You Roq! :)

PR #536, opened 2025-06-09
Adds diagram plugin

jtama:
Adds diagram plugin using kroki.io to generate images

### DIFF
--- a/blog/content/docs/plugins.adoc
+++ b/blog/content/docs/plugins.adoc
@@ -311,3 +311,58 @@ layout: roq-default/default
 search: false
 ---
 ----
+
+[#plugin-diagram]
+== Roq Plugin Diagram
+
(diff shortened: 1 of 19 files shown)


jtama:
@melloware thanks for rebasing !

jtama:
Ok so now the blog depends on a container which is apparently not available for windows env so 3 solutions : 

1. Rely on the kroki.io server
2. Amend the CI so we have a container runtime on the window runner (from far my favourite)
3. Remove the plugin usage from the blog

melloware:
OK lets remove plugin usage from the blog for now since it can't build on Windows!

jtama:
done

jtama:
Ok I guess I know what it is, the need for the devservice is detected, so it tries, let me juste remove the dep.

melloware (approved):

PR #559, opened 2025-07-15
Update advanced.adoc to fix #532

### LINKED ISSUE #532: RoqAndRoll is configured to port 8082 by default, but that's not documented
Usually QuarkusTest runs on Quarkus using port 8081 by default and test run logs showing there is also an instance of Quarkus running on port 8081, but when the test is annotated with RoqAndRoll, it runs an additional instance on port 8082 because of it's (interface) default implementation.

I was testing some Microservices and already running one on 8082, so the port was blocked and the tests running against the existing service, where the content check fails, but not the Test itself because of the blocked port...
That was unexpected.

Workaround was to configure a different port, like this `@RoqAndRoll(port=8083)`.

At minimum, it might be good to document that port and configuration.
I could create a PR for that, if you like to.

But further examination turned out, that usual QuarkusTest runs (without `@RoqAndRoll` annotation) working with the instance running on port 8081 too (!)
So in general, why that specific annotation for Roq testing is required at all and not regular Quarkus testing could be used on port 8081 only, especially when there is an additional instance of Quarkus running on that port already?

janwesterkamp:
Refactorings and additions to document Roq test support.

### DIFF
--- a/blog/content/docs/advanced.adoc
+++ b/blog/content/docs/advanced.adoc
@@ -476,9 +476,32 @@ public class RoqSiteTest {
 }
 ----
 
-The RestAssured port will automatically use the Roq static server, running on port 8082 by default.
+The RestAssured port will automatically use the Roq static test server, running on port 8082 by default.
 
 TIP: The Roq test server port could be modified by an annotation parameter like this `@RoqAndRoll(port=8082)`.
+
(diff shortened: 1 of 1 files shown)


janwesterkamp:
This will fix https://github.com/quarkiverse/quarkus-roq/issues/532.

PR #571, opened 2025-07-24
Introduce Fragment search

ia3andy:


roq-plugin/lunr/runtime/src/main/java/io/quarkiverse/roq/plugin/lunr/runtime/RoqPluginLunrTemplateExtension.java
```diff
@@ -0,0 +1,110 @@
...
+        Elements withId = html.select("[id]");
+        List<Anchor> anchors = new ArrayList<>();
+
+        for (Element el : withId) {
+            String id = el.id();
+            String title;
+            String content;
+
+            if (el.tagName().matches("h[1-6]")) {
+                // For headings, use parent text as content
```
jtama: Why would you use the parent as content ? Seems strange to me.

jtama (commented):
Overall code and experience is great.

One warning though. I think it's fine tuning, but the boost on fragment is huge, and for our blog, we add another boost for the documentation, which sends the blog posts at the very end of the queue.

With the actual tuning and code, a 3 level title on documentation weights 15 more than a blog title.

roq-plugin/lunr/runtime/src/main/java/io/quarkiverse/roq/plugin/lunr/runtime/RoqPluginLunrTemplateExtension.java
ia3andy: That's a very good question :) I suppose this works only if there a no titles before that one.. this is how asciidoc does it:
```html
div.section 
  h2#foo
  content

File: blog/templates/layouts/doc.html [hunk 2]
```diff
@@ -1,6 +1,7 @@
 ---
 layout: :theme/main
 bodyClass: documentation
+index-boost: 30
```
ia3andy: @jtama isn't that what you mean by parameterised boost? With this you can choose what you want to appear first in the search.

PR #578, opened 2025-08-01
fix slugify path when file hasn't got any extension.

### LINKED ISSUE #577: Slugify adds unexpected `null` string when file path doesn't contains extension

jtama:
Fixes #577


ia3andy (approved):
LGTM!

PR #584, opened 2025-08-04
feat: configurable list of social brands via data

jotaNas:
This PR implements a configurable list of social brands in the sidebar-contact.html using the @TemplateGlobal strategy, as suggested in the issue discussion.

Each brand's configuration (id and icon) is now defined in a global Java class, and the template dynamically builds the correct URL using site.data.

This avoids hardcoding HTML blocks and allows easier customization via site.data.


ia3andy:
This looks very much like a PR implemented by AI :)

ia3andy:
It's too far from the right implementation to be reviewable.. (it would take more time to review rather than implement it)

PR #597, opened 2025-08-07
Start clean for the generated config doc

gsmet:
Also drop the @ConfigDocValue as it doesn't fix the issue and is going to be annoying to maintain.

The issue will be properly fixed in the next Quarkus release and the doc will have to be generated again.

(If you absolutely want to work around the issue now, you can disable formatting by surrounding the value with `+++` (that's what the actual fix will do)


ia3andy (approved):
Thanks @gsmet :)

PR #615, opened 2025-08-26
Update plugins.adoc to qualify asciidoc-java implementation

rmannibucau:
Avoid ambiguities since there is no direct link with asciidoc org/projects except the syntax


jtama (approved):
Yeap, looks good.

Issue #650, opened 2025-11-05
i18n Support


ia3andy:
- [ ] enable overridable theme messages Quarkus 3.30+
- [x] allow to define locale that will be used by Qute for rendering via FM #657 
- [ ] allow to create i18n collections #680  

### Language-Specific Pages in Collections

**Goal**  
Pages inside a collection directory should behave like attachments to the parent document, not standalone posts.

**Proposed Behavior**  
- Files like `fr.adoc` should be accessible at `<parent-doc-url>/fr/` instead of being published as independent documents.
- No need for derived collections; pagination stays on the existing collection.

**Technical Notes**  
- Update detection logic in `RoqFrontMatterScanProcessor` so these pages are treated as normal pages, not posts.  
  [Relevant code](https://github.com/quarkiverse/quarkus-roq/blob/main/roq-frontmatter/deployment/src/main/java/ioon with `hidden`, `draft`, and `future` flags.
- Optional: sub-collection approach, but likely unnecessary.

**Open Questions**  
- Should language pages inherit metadata from the parent?
- Naming convention vs explicit config?

matheusandre1:
hey @ia3andy and @jtama Can I propose a PR for this? I have some free time to work on this issue.

PR #657, opened 2025-11-12
Allow to define the locale that will be used by Qute for rendering.

jtama:
In ascending order of importance :

1. The value of `quarkus.default-locale`
1. The value of `site.default-locale`
1. The value of the accept-header locale
1. The value of the FM `locale` attribute

See #650


ia3andy (approved):
Thanks!

Issue #773, opened 2026-03-05
Editor no longer starts


edewit:
The roq editor no longer loads but instead gives this error in the console:

```
Uncaught (in promise) TypeError: The specifier “@tiptap/core” was a bare specifier, but was not remapped to anything. Relative module specifiers must start with “./”, “../” or “/”.
```

It seems that the bundle isn't being created

PR #774, opened 2026-03-05
added missing dependencies and exclude to prevent overwrite

edewit:
fixes: #773
Signed-off-by: Erik Jan de Wit <erikjan.dewit@gmail.com>

### DIFF
--- a/roq-editor/deployment/pom.xml
+++ b/roq-editor/deployment/pom.xml
@@ -50,6 +50,14 @@
     </dependencies>
 
     <build>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <excludes>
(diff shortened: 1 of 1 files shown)


roq-editor/deployment/pom.xml
```diff
@@ -50,6 +50,14 @@
     </dependencies>
 
     <build>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <excludes>
+                    <exclude>dev-ui/bundle.js</exclude>
```
ia3andy, line 57: @edewit why do we need to exclude it?

edewit: it overwrites the one that was build by the esbuild maven plugin

PR #776, opened 2026-03-09
Add TOC plugin for server-side table of contents generation

### LINKED ISSUE #579: Create a TOC plugin
It would allow to have a nice dynamic TOC like this: 

Image

For Markdown and Asciidoc content.

This is generated from JS, we could improve it by generating the TOC dom in the content.

rolfedh:
## Summary

* Fixes #579
* Adds a new `roq-plugin-toc` module that generates a table of contents from page headings in Markdown and AsciiDoc content
* The plugin is a standalone opt-in dependency (not bundled in the `quarkus-roq` aggregate extension)

## How it works

The plugin parses rendered page HTML with JSoup, extracts headings (`h1`–`h6`) that have `id` attributes, and builds a nested hierarchy. It exposes two Qute template extensions on `Page`:

- **`{page.toc}`** — returns a `List<TocEntry>` with `id`, `title`, `level`, and `children` properties for custom rendering
- **`{page.tocHtml}`** — returns a pre-rendered `<nav class="roq-toc">` HTML block with nested `<ul>` lists and anchor links

Both AsciiDoc (`div.sect*` sections) and Markdown (`h*[id]` elements) heading formats are supported. All HTML output is XSS-escaped.

This implementation generates static HTML on the server side, consistent with the issue request to "generate the TOC dom in the content." JavaScript-based scroll tracking or active-section highlighting would be a follow-up enhancement.

## What's included

- `roq-plugin/toc/` — new plugin module (deployment + runtime)
- 16 unit tests covering heading extraction (Markdown + AsciiDoc), hierarchy building, HTML rendering, and XSS escaping
- AsciiDoc documentation page with concept, how-to, and reference sections
- Navigation and index updates

## LLM disclosure

This PR was developed with assistance from Claude Code (Claude Opus 4.6). All code was reviewed, tested, and verified by the author.

## Test plan

- [x] `mvn install -DskipTests` — full project build succeeds
- [x] `mvn test -pl roq-plugin/toc/runtime` — 16 unit tests pass
- [x] `mvn verify` — no cross-module regressions (pre-existing `RoqCodestartTest` failures unrelated)
- [x] Manually tested with `blog/` example: TOC renders correctly for both Markdown and AsciiDoc posts


rolfedh:
@ia3andy, you can disposition this PR whenever you're ready. Thanks!

rolfedh:
@ia3andy Quick follow-up: just pushed c52d3ec, which makes #776 easier to evaluate alongside #770 and #797:

- `{page.tocHtml}` now emits `<nav aria-label="…">` with `data-level` on each `<li>` (0-indexed depth, matching the default theme's `toc.js`) — so the existing JS TOC can decorate the server-rendered output via progressive enhancement.
- Honors the same frontmatter keys the default theme uses: `content-toc` (opt-out), `content-toc-levels` (max depth), `content-toc-title` (label override).
- Content-hash-keyed cache avoids re-parsing unchanged pages.

All 27 unit tests pass locally; PR is still `MERGEABLE` / `CLEAN`.

**TL;DR:** a server-rendered TOC gives AI crawlers a parseable page skeleton — a per-page complement to the site-level `llms.txt` index from #797 — without taking anything away from the JS TOC. If it's not a fit, happy to close the PR; no hard feelings either way.

rolfedh:
Closing this to keep the queue tidy. The JS-based TOC that shipped in the default theme with 2.1.0 covers the common case well, and I don't want a parallel server-side variant lingering unreviewed.

If there's ever appetite for a no-JS / build-time TOC (accessibility, SEO, no-script rendering), I'm happy to revisit or extract it as a standalone plugin — just give me a ping. Thanks @ia3andy!

holly-cummins:
I think @rolfedh is right, and we will need a TOC plugin to provide TOC support for sites which do not use the default theme. I noticed this when doing the Jekyll conversion for quarkus.io (and I imagine @rolfedh noticed it for similar reasons). We may be able to get away without a toc plugin for quarkus.io specifically, but there will be a need for it in general.

jtama:
I would really like this one.

ia3andy:
I think the client side one is better we could write a tutorial pointing to the default theme js

jtama:
Even if it's client side, it would be easier to carve it out in a plugin, so users don't have to copy/paste code, no ?

holly-cummins:
> I think the client side one is better we could write a tutorial pointing to the default theme js

What's the reason for preferring client side? My gut feeling is that if you have a static site generator, it's a perfect opportunity to do things like introspect the site and work out a table of contents, rather than adding cpu cycles on the client side (especially since if it's server side it could render early, but client-side I think it would have to be one of the last things to render). 

But either way, I think it's nicest to have it available as a plugin without requiring people to redo the work of extracting it from the default theme.

holly-cummins:
This is needed for quarkus.io, so I've pulled across the code from Rolfe's branch and built it in quarkus.io: https://github.com/quarkusio/quarkusio.github.io/pull/2929 as a temporary solution

roq-plugin/toc/runtime/src/main/java/io/quarkiverse/roq/plugin/toc/runtime/RoqPluginTocTemplateExtension.java
```diff
@@ -0,0 +1,303 @@
...
+            if (!entry.children().isEmpty()) {
+                sb.append('\n');
+                renderEntries(sb, entry.children());
+            }
+            sb.append("</li>\n");
+        }
+        sb.append("</ul>\n");
+    }
+
+    static String escapeHtml(String text) {
```
holly-cummins:
```suggestion
    private static String escapeHtml(String text) {
```

roq-plugin/toc/runtime/src/main/java/io/quarkiverse/roq/plugin/toc/runtime/RoqPluginTocTemplateExtension.java
```diff
@@ -0,0 +1,303 @@
...
+        return Entities.escape(text, new Document.OutputSettings().charset("UTF-8").escapeMode(Entities.EscapeMode.base));
+    }
+
+    static String escapeAttr(String text) {
```
holly-cummins:
```suggestion
    private static String escapeAttr(String text) {
```

If these methods aren't private, Qute registers them as template extensions and it can cause interference elsewhere.

roq-plugin/toc/runtime/src/main/java/io/quarkiverse/roq/plugin/toc/runtime/RoqPluginTocTemplateExtension.java
holly-cummins: Ignore these comments, I see you solved this in a different way

roq-plugin/toc/runtime/src/main/java/io/quarkiverse/roq/plugin/toc/runtime/RoqPluginTocTemplateExtension.java
holly-cummins: Ignore these comments, I see you solved this in a different way!

cescoffier:
+1 to move it into a plugin, makes sense.

About build time vs. JavaScript, I would be on the build-time side. Use every DOM we know at build time. JavaScript makes the page less consumable by crawlers (like AI agents), increases CPU consumption on the client side, and the "win" in terms of page size is meaningless (especially because for large documents, the DOM parsing to do in JavaScript will allocate so many objects that the memory used in the browser will be quite bad).

rolfedh:
Update on the AsciiDoc `toclevels` support, since the PR description predates it.

`{page.toc}` and `{page.tocHtml}` now follow the source document's TOC depth instead of always emitting every heading level:

- `content-toc-levels` in front matter still wins when a page sets it.
- Otherwise, on AsciiDoc pages, the plugin uses the document's `toclevels` attribute and adds one, because `toclevels` counts section depth while the plugin filters on heading tag level. It looks in the `asciidoc-attributes` front matter map, then in the attributes Roq parses from the document header, then at a `:toclevels:` entry in the header itself.
- With neither set, AsciiDoc pages show two section levels, which is Asciidoctor's own default. Other pages still show all six.

The latest commit fixes the attribute lookup. It read only the `asciidoc-attributes` key, which a page sets for itself, and not `attributes`, where Roq files what it parses from a document header, so a parsed header never supplied the value. Reading both also picks up a `toclevels` set in an included attributes file, which a scan of the page source cannot see.

The header scan is now bounded to the document header with the last entry winning, matching how Asciidoctor resolves attribute entries, and accepts the soft-set form `:toclevels: 3@`.

Verified on quarkus.io, which builds this plugin from the branch: a full 5341-page build gives logging four levels, qute-reference three, getting-started two, cdi-integration two, and the OIDC code flow guide three, each matching Asciidoctor's own `:toc:` outline for the same source. 44 unit tests pass.

rolfedh:
One piece of housekeeping before this merges, and it is my oversight rather than anything in the code.

The commit `d30c3175` on this branch ends with a `Co-Authored-By: Claude Opus 5` trailer, which CONTRIBUTING.md asks contributors not to use.

Two ways to clear it, whichever is less work for you:

- Drop the trailer when you squash-merge. The squash message is composed at merge time anyway, so this costs you nothing extra.
- Or I rewrite the branch to remove it. One thing to know if I do: quarkus.io currently builds this plugin from a pinned commit on this branch (https://github.com/quarkusio/quarkusio.github.io/pull/2936), so rewriting changes that SHA and I would repin the site PR in the same pass. Happy to do it, just say the word.

Either way it does not block anything, and no rush.

rolfedh:
@holly-cummins FYI

docs/modules/ROOT/pages/quarkus-roq-plugin-toc.adoc
ia3andy, line 1: @rolfedh plugin doc should be in the blog marketplace instead (following the others)

ia3andy (changes requested):
One concern I have, I am not sure how to fix it is if someone use page.toc inside the page content itself...

rolfedh:
@ia3andy Good catch. I traced what happens today: toc() calls site.pageContent(page), which renders the page content on its own. When the content itself contains {page.toc} or {page.tocHtml}, that nested render re-enters toc() for the same page. Roq's guard on Page.content() already cuts the loop, so the output was correct, but it logged "WARN Recursive call to {page.content} detected" on every render of such a page.

Fixed in baed0622: the extension now tracks the pages it is resolving on the current thread and returns an empty list on re-entry, so the nested render stays finite and nothing is logged at WARN. The outer call still returns the real entries, and the inline TOC in the final page is populated too, because page.content() is memoized.

The new RoqPluginTocInContentTest renders a page with {page.tocHtml} inside its content, asserts that both the layout TOC and the inline TOC list the headings, and fails on any WARN record. I confirmed it fails without the guard, with exactly that recursion warning, and passes with it.

Two more things in the same push: the docs moved to the marketplace as you asked (07015acc), and I rewrote the branch to drop the two Co-Authored-By trailers per CONTRIBUTING.md, so every SHA changed. quarkus.io pins the plugin by SHA, so a repin PR follows there.

Repin PR on quarkus.io: https://github.com/quarkusio/quarkusio.github.io/pull/2945

docs/modules/ROOT/pages/quarkus-roq-plugin-toc.adoc
rolfedh: Done in 07015acc: the plugin is now documented at blog/content/marketplace/plugins/toc.md, following the other plugin pages, and the Antora page with its nav and index entries is gone. It will show at https://quarkiverse-roq-blog-776-preview.surge.sh/plugin/toc/ once the PR preview rebuilds.

One thing I noticed while there, not part of this PR: docs/modules/ROOT/nav.adoc on main still links quarkus-roq-plugin-og-card.adoc, and that page no longer exists in docs/modules/ROOT/pages since #1046.

roq-plugin/toc/runtime/src/main/java/io/quarkiverse/roq/plugin/toc/runtime/RoqPluginTocTemplateExtension.java
rolfedh: Closing the loop, since the thread is still open. Both halves are now on the branch: 7f37c765 moved `@TemplateExtension` from the class onto the two methods it belongs on, `toc` and `tocHtml`, so Qute registers only those, and 5007c9e7 also made `escapeHtml` and `escapeAttr` private as you suggested, so a future class-level annotation could not re-expose them. Your concern was real: the class-level registration is what corrupted the RSS feed in quarkusio/quarkusio.github.io#2929, and `RssFeedTest` there now pins it. Fine to resolve.

roq-plugin/toc/runtime/src/main/java/io/quarkiverse/roq/plugin/toc/runtime/RoqPluginTocTemplateExtension.java
rolfedh: Same as the `escapeHtml` thread: private since 5007c9e7, and only `toc` and `tocHtml` carry `@TemplateExtension` since 7f37c765. Fine to resolve.

rolfedh:
@ia3andy One piece of context for scheduling: quarkus.io merged the temporary build of this plugin on 25 August and has served the guide table of contents from it since, and Guillaume Smet asked on the quarkus.io side (https://github.com/quarkusio/quarkusio.github.io/pull/2945#issuecomment-5426384656) to replace that with the released plugin soon. Everything from your review is on the branch and CI is green, so whenever you merge, a release soon after would let us drop the workaround.

ia3andy (approved):
LGTM

rolfedh:
Thanks @ia3andy — much appreciated, and for getting it into 2.1.8 right away!

PR #777, opened 2026-03-10
Fix include directives for non-AsciiDoc files

rolfedh:
## Summary

- Add configurable `quarkus.asciidoc.include-extensions` property to control which non-AsciiDoc file types the custom include processor handles
- Default: `.java,.kt,.js,.ts,.groovy,.scala` (languages using `//`-style tag markers)
- `.adoc`/`.asciidoc` files are always handled regardless of config
- Fix URL prefix check (`matches()` → `find()`) so URL include targets are correctly delegated to the default handler

## Problem

The custom `AsciidocJInclude` processor's `handles()` method only accepted `.adoc` and `.asciidoc` targets. Include directives for other file types fell through to AsciidoctorJ's default include handler.

For **co-located files** (same directory), the default handler works fine. But for **cross-directory includes** using document attributes, the default handler fails because Roq feeds content to AsciidoctorJ via `asciidoctor.load()` (as a string, not from a file), so there is no source file context to resolve relative paths against.

For example, this pattern fails with the default handler:
```asciidoc
[source,java]
----
include::{generated-dir}/examples/MyApp.java[tags=example]
----
```

## Fix

Add a configurable allowlist of file extensions (`quarkus.asciidoc.include-extensions`) that the custom processor handles. The default list covers languages where `//`-style tag markers work correctly. Users can extend or restrict the list via `application.properties`.

The existing path resolution, security boundary enforcement (ROOTDIR check), and tag/line extraction logic already work correctly for all file types.

## Test plan

- [x] All 15 tests pass (9 existing + 6 new `AsciidocJIncludeHandlesTest` for `handles()` logic)
- [x] Blog example builds successfully (`mvn verify -pl blog`)
- [x] Verified with quarkusio.github.io guides that use `include::` for `.java` files with tag extraction

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ia3andy (changes requested):
I would feel safer if we had a list of authrized extensions for include in the config

jtama (approved):
LGTM

roq-plugin/asciidoc-jruby/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/runtime/AsciidocJInclude.java
```diff
@@ -122,11 +126,11 @@ static List<String> processInclude(String content, Map<String, Object> attrs) th
...
 
     // Extract lines between tag::...[] and end::...[]
+    // Supports standard AsciiDoc comment styles: //, #, --, ;;, <!--
     private static List<String> extractTag(List<String> lines, String tag) {
-        // Ruby: start = /^\/\/\s*tag::#{tag}\[\]/
-        //       end   = /^\/\/\s*end::#{tag}\[\]/
-        Pattern start = Pattern.compile("^//\\s*tag::" + Pattern.quote(tag) + "\\[\\]");
-        Pattern end = Pattern.compile("^//\\s*end::" + Pattern.quote(tag) + "\\[\\]");
+        String quotedTag = Pattern.quote(tag);
+        Pattern start = Pattern.compile("^(?://|#|--|;;|<!--)\\s*tag::" + quotedTag + "\\[\\]");
```
jtama: One last detail, those patterns should be compiled once and for all.

jtama (changes requested):

rolfedh (commented):
@jtama @ia3andy Addressed the feedback — `TAG_START_PATTERN` and `TAG_END_PATTERN` are now compiled once as `static final` fields with a capturing group, and the tag name is matched via `.equals()` instead of being baked into the regex.

jtama (approved):

ia3andy (approved):
thanks :)

PR #862, opened 2026-04-27
Allow to generate page from content if specified by user.

jtama:
The PR allow user to trigger page generation from roq-data.

The user must explicitely ask for it and provide collection layout and key attribute.

Pages will be empty except from frontmatter data and redering is to be done through layout.


roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep2AssembleProcessor.java
```diff
@@ -112,4 +127,77 @@ void processLayouts(RoqSiteConfig config,
                     processed.generatedTemplate(), isThemeLayout));
         }
     }
+
+    // ── Roq Data processing ───────────────────────────────────────────────
+
+    @BuildStep
+    void processContent(RoqProjectBuildItem roqProject,
```
ia3andy:
```suggestion
    void processDataContent(RoqProjectBuildItem roqProject,
```

roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep2AssembleProcessor.java
```diff
@@ -112,4 +127,77 @@ void processLayouts(RoqSiteConfig config,
...
+            }
+
+        }
+    }
+
+    private void generatePages(String collection, RoqDataNamedConfig collectionConfig, JsonObject item,
+            BuildProducer<RoqFrontMatterRawPageBuildItem> rawPagesProducer, RoqProjectBuildItem roqProject) {
+        final String extractedKey = item.getString(collectionConfig.titleAttributeName());
+        if (extractedKey == null) {
+            throw new IllegalStateException("Extracted key for % is null with property key %".formatted(item.toString(),
```
ia3andy: Use existing or new exception with ` extends RoqException`

roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep2AssembleProcessor.java
```diff
@@ -112,4 +127,77 @@ void processLayouts(RoqSiteConfig config,
...
+                case JsonArray jsonArray ->
+                    jsonArray.forEach(jsonObject -> generatePages(itemName, collectionConfig, (JsonObject) jsonObject,
+                            rawPageProducer, roqProject));
+                default -> throw new IllegalStateException();
+            }
+
+        }
+    }
+
+    private void generatePages(String collection, RoqDataNamedConfig collectionConfig, JsonObject item,
```
ia3andy:
```suggestion
    private void generateDataPages(String collection, RoqDataNamedConfig collectionConfig, JsonObject item,
```

roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep2AssembleProcessor.java
```diff
@@ -112,4 +127,77 @@ void processLayouts(RoqSiteConfig config,
...
+                item,
+                new ConfiguredCollection(collection, false, false, false, collectionConfig.layout()),
+                getIncludeFilter(layoutId).apply(""),
+                "",
+                List.of()));
+    }
+
+    private Object convert(DataMappingBuildItem dataMappingBuildItem) {
+        try {
+            return dataMappingBuildItem.getConverter().convert(dataMappingBuildItem.getContent());
```
ia3andy, line 200: Maybe we should have a new builditem in data, which hold the converted result? This way you don't have to convert twice.

roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep2AssembleProcessor.java
```diff
@@ -112,4 +127,77 @@ void processLayouts(RoqSiteConfig config,
...
+        } catch (IOException e) {
+            throw new DataConversionException(
```
ia3andy, line 202: We should move RoqException to common, so we use it in data too I suppose?

ia3andy (changes requested):

roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep2AssembleProcessor.java
jtama: Not sure I understand. The bean can be converted to a typed instance if there's a @DataMapping somewhere, so both conversion is needed in this case.

You might have something else in mind.

roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep2AssembleProcessor.java
jtama: Nope `DataReadingException` is used  there [RoqDataConverterProcessor](https://github.com/quarkiverse/quarkus-roq/blob/main/roq-data/deployment/src/main/java/io/quarkiverse/roq/data/deployment/RoqDataConverterProcessor.java#L66)

ia3andy: ok I need to add support for Nice Data exception in frontmatter

roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/config/RoqSiteConfig.java
```diff
@@ -249,12 +255,29 @@ interface CollectionConfig {
...
          * Resolves local layout first, then theme layout as fallback.
          */
         Optional<String> layout();
+
+        /**
+         * The attribute to be used as title
+         *
+         * @asciidoclet
+         */
+        Optional<String> titleAttributeName();
```
ia3andy: I would prefer if we keep this in `Optional<CollectionFromData> generateFromData` as `interface CollectionFromData` with `idKey` and `Optional<String>` dataName()

ia3andy: The reason is that `titleAttributeName` doesn't make sense when not generating from data

jtama: Ok the then maybe I can change the logic. Keep the optional (rename to idKey), if present, then it means pages are to be generated, and I drop the `generateFromData` boolean ?

roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/config/RoqSiteConfig.java
```diff
@@ -274,10 +264,18 @@ interface CollectionConfig {
         Optional<String> layout();
 
         /**
-         * The attribute to be used as title
+         * If present, a page will be created for each item in this collection
```
ia3andy: "If present, documents for this collection will be created for the data (array or dir) with the same name as the collection (using the default layout)"

roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/config/ConfiguredCollection.java
```diff
@@ -9,12 +9,17 @@ public record ConfiguredCollection(
         boolean future,
         String layout,
         boolean generate,
-        String titleAttributeName) {
+        String idKey) {
```
ia3andy: keep the CollectionFromData here, you can create a record for it if needed

roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep2AssembleProcessor.java
```diff
@@ -112,4 +129,78 @@ void processLayouts(RoqSiteConfig config,
...
+            BuildProducer<RoqFrontMatterRawPageBuildItem> rawPagesProducer, RoqProjectBuildItem roqProject) {
+        final String extractedKey = item.getString(configuredCollection.idKey());
+        if (extractedKey == null) {
+            throw new RoqFrontMatterReadingException(RoqException
+                    .builder("Error extracting title value from %s collection values".formatted(configuredCollection.id()))
+                    .hint("Extracted key for %s is null with property key %s".formatted(item.toString(),
+                            configuredCollection.idKey())));
+        }
+        final String id = configuredCollection.id() + "/" + slugify(extractedKey, false, false);
+        final String path = id + ".md";
```
ia3andy: why .md an not .html?

PR #899, opened 2026-05-09
Add Prism syntax highlighting plugin

cdhermann:
## Summary

Adds a new Roq plugin, `quarkus-roq-plugin-prism`, that provides self-hosted [Prism](https://prismjs.com/) syntax highlighting for code blocks. It features a small footprint and removes the need for a CDN.

> [!NOTE]
> Although the output is JavaScript, we don't use `io.mvnpm:esbuild-maven-plugin`. Prism's per-component files are pre-minified and designed to concatenate, so we read them straight from the `org.mvnpm:prismjs` JAR and assemble the bundle in a build step — no Node toolchain needed.

## Usage

Add the plugin to a Roq site:

```xml
<dependency>
    <groupId>io.quarkiverse.roq</groupId>
    <artifactId>quarkus-roq-plugin-prism</artifactId>
    <version>${quarkus-roq.version}</version>
</dependency>
```

Configure which languages to bundle and (optionally) which theme in `application.properties`:

```properties
quarkus.roq.prism.languages=java,bash,properties,yaml,markup-templating,json
quarkus.roq.prism.theme=tomorrow
```

| Property | Description | Default |
| --- | --- | --- |
| `quarkus.roq.prism.languages` | Canonical names from Prism's [components.json](https://github.com/PrismJS/prism/blob/master/components.json). Transitive `require` prereqs are resolved automatically. | *required* |
| `quarkus.roq.prism.theme` | One of `default`, `coy`, `dark`, `funky`, `okaidia`, `solarizedlight`, `tomorrow`, `twilight`. | `default` |

Insert the `{#prism /}` Qute tag in your base layout's `<head>` to emit the `<link>` and deferred `<script>` for the bundle:

```html
<head>
  ...
  {#prism /}
  {#bundle /}
</head>
```

The script auto-runs `Prism.highlightAll()` once the DOM is parsed, so any `<pre><code class="language-X">…</code></pre>` produced by your Markdown/AsciiDoc renderer gets highlighted with no further JS on your part.

*Still to be examined is how users can activate syntax highlighting in the frontmatter without having the need to modify the template code. <https://github.com/itbh-at/quarkus-roq/tree/feat/roq-head-contributor> contains an implementation idea but introduces an HTML head contributor component.*

## Design

At build time, `RoqPluginPrismProcessor` reads `prism-core.min.js` + each requested language's `prism-<name>.min.js` from the `org.mvnpm:prismjs` JAR, concatenates them with an auto-highlight trigger, and emits the result alongside the chosen theme as `GeneratedStaticResourceBuildItem`s at `/static/bundle/prism.{js,css}`. No Node toolchain; the consumer ships only the languages they configured.

> [!IMPORTANT]
> I am still not sure if the idea of having global highlighting support for any input language matches the intent of Roq since AsciiDoc as input language seems to feature its own syntax highlighting using highlight.js


ia3andy:
Why not just using prism light we do with highlightjs? 
https://iamroq.dev/posts/how-to-add-syntax-highlighting-to-your-roq-site/#step-2-initialize-highlightjs

cdhermann:
> Why not just using prism light we do with highlightjs?
> https://iamroq.dev/posts/how-to-add-syntax-highlighting-to-your-roq-site/#step-2-initialize-highlightjs

There are several reasons for this approach:

1. **Curiosity**: It provided a great opportunity to explore and understand the plugin mechanism.

2. **Performance & Architecture**: From a user's perspective, enabling a plugin is generally much more natural than injecting custom JavaScript. While syntax highlighting implies some programming knowledge, the underlying principle applies to other features too. For example, with automated image format conversion (as demonstrated [here](https://www.itbh.at/posts/automated-webp-conversion-in-a-roq-static-site/)), it is much cleaner to use a build-time plugin than to write client-side JS that replaces image URLs at runtime. The overall goal is to handle these tasks at build time to keep the runtime overhead in the browser as minimal as possible.

3. **Bandwidth**: Prism.js is favored over Highlight.js due to its significantly smaller footprint.

4. **Module Compatibility**: I am not familiar with "prism light" and whether it is an ES module. However, standard Prism.js is not natively distributed as an ES module, which is - according to my understanding- why it cannot be integrated in the exact same way as Highlight.js.

ia3andy:
I am not convinced by your AI generated arguments. We have the web bundler in Roq/Quarkus we don't need to re-invent the wheel to have js libraries. I am not against a plugin but it should be way way simpler.

PR #913, opened 2026-05-19
Disable editor auto-sync file name by default

ia3andy:
## Summary
- Change the default value of `editor.suggestedPath.enabled` from `true` to `false`
- Users can still opt-in by setting `editor.suggestedPath.enabled=true` in their configuration


ia3andy:
I like this feature (I made it) but with long title, it's really a pain :) I need to find a way to make it better, maybe truncate the size for the file name or have a new "short" slug method

PR #947, opened 2026-05-26
feat(editor): improve sync status bar UX

### LINKED ISSUE #914: Editor Sync: improve status bar UX and visual integration
Follow-up from #768.

## Visual integration

- The sync status bar should be more visually integrated with the editor header (not a separate row).
- Place the status bar on the **right** side of the header, next to the save/create buttons.

### Current layout (to improve)

Image

Image

Image

## Status bar behavior

- Don't blink/pulse the icon during sync. Instead, spin the checkmark (V) icon while syncing.
- When synced: show **"Git is synced"** with a 🔁 refresh button next to it (no label but popover).
- When there are local changes: show **"New content to publish"** with a publish icon (instead of the sync icon).
- Hide the publish button entirely when everything is synced (no need to show it if there's nothing to publish).

## Summary

The goal is a cleaner, more compact layout where status + actions live in one integrated bar, with contextual buttons that only appear when relevant.

cc @omatheusmesmo

Ivictors:
Integrate the sync status bar into the editor and content headers instead of rendering it in a separate row.

Use contextual icon-only actions, spin the checkmark during sync, and remove the dedicated sync controls component.

Align the status text and button visibility with the UX requested in the issue while keeping the changes scoped to the Dev UI frontend.

Closes #914


ia3andy (commented):
Thanks for this PR! The consolidation of sync controls into the status bar is a nice cleanup. A few things I noticed:

**1. Missing `config.sync?.enabled` guard** (`base-editor.js`, `pages-list.js`)
The old code checked `config.sync?.enabled && isGitRepo` before rendering sync UI. Now only branch presence is checked via `hasSync`, so the sync status bar will appear even when sync is disabled in config.

**2. Spinning checkmark is misleading** (`sync-status-bar.js`)
The syncing state now shows a spinning `fa:check` icon. A checkmark implies "done/success", so spinning it sends a contradictory signal. The old `fa:rotate` icon was the conventional choice for "in progress".

**3. No visual feedback during publish** (`sync-status-bar.js`)
`_renderStatus()` only checks `this.syncing` for the spinner, never `this.publishing`. During a publish operation, there's no spinner feedback for the user.

**4. `_hasSyncAction()` always true when synced** (`sync-status-bar.js`)
Including `upToDate` in the condition means the refresh button never disappears. Is that intentional (allow manual re-check), or should it only show when there's actually something to pull?

ia3andy:
image

ok it's getting there, still a few things to make it perfect:
- search should stay centered
- github bar should have a light gray background and radius to be visually appearing (the tag could have the orange bg from the cloud
- actions icons shouldn't be ugly
- the refresh icon should be the check mark on mouse hover
- make it look nice overall

ia3andy (approved):
thanks @Ivictors please do the follow up PR based on my comment

PR #951, opened 2026-05-30
Add slugify case control

### LINKED ISSUE #895: Expose link placeholder values as template extensions and add slugify options
## Problem

Link placeholders (`:slug`, `:name`, `:path`) are computed in `TemplateLink.java` during URL generation but not accessible from Qute templates. This makes cross-collection linking fragile since template authors must manually replicate the slug resolution logic.

Additionally, the `.slugify` template extension has no options. Internally `StringPaths.slugify(text, preserveCase, preservePath)` supports case control, but the template method always calls `slugify(text, false, false)`. The URL generation itself uses `.toLowerCase()` after slugify, so there's no way for templates to produce the same result.

### 1. Add page link placeholder extensions

Expose template extension methods on `Page` that match what `TemplateLink` resolves:

- `page.slug` / `page.Slug`
- `page.name` / `page.Name`

### 2. Add options to `.slugify`

Allow controlling case in the template extension:

```html
{page.title.slugify}           {! current behavior, case-preserving !}
{page.title.slugify('lower')}  {! lowercased !}
```

## Use case

Two collections share the same titles (e.g. `digest-posts` and `digest-swipe`). To link from a post to its swipe page:

```html
{! Today: not possible without manually replicating the URL generation logic !}

{! With page.slug: !}
<a href="{site.url('digest-swipe', page.slug)}">Swipe mode</a>
```

matheusandre1:
Closes: #895

### DIFF
--- a/roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/utils/TemplateLink.java
+++ b/roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/utils/TemplateLink.java
@@ -80,34 +80,41 @@ public class TemplateLink {
                         () -> data.pageSource().isTargetHtml() ? ".html" : "." + data.pageSource().extension()),
                 Map.entry(":slug", () -> resolveSlug(data).toLowerCase()),
                 Map.entry(":Slug", () -> resolveSlug(data)),
-                Map.entry(":name", () -> StringPaths.slugify(resolveName(data), true, false)
-                        .toLowerCase()),
-                Map.entry(":Name", () -> StringPaths.slugify(resolveName(data), true, false)))); // Case-preserving slug
+                Map.entry(":name", () -> resolveName(data).toLowerCase()),
+                Map.entry(":Name", () -> resolveName(data))));
         if (other != null) {
             result.putAll(other);
         }
         return result;
     }
 
-    private static String resolveName(LinkData data) {
+    public static String resolveName(PageSource pageSource) {
         final String name;
-        if (data.pageSource().isIndex()) {
-            final Path parent = Path.of(data.pageSource().path()).getParent();
+        if (pageSource.isIndex()) {
+            final Path parent = Path.of(pageSource.path()).getParent();
             name = parent != null ? parent.getFileName().toString() : "not-available";
         } else {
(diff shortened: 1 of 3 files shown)


ia3andy (commented):
--

ia3andy (commented):
---

ia3andy:
(Sorry for the duplicate review above, please ignore them.)

**TL;DR of the requested changes:**

1. **Keep only** `page.slug`, `page.Slug`, `page.name`, `page.Name`. Remove `pathSlug` and `rawPath`.
2. **Avoid exposing `LinkData`**: refactor `resolveSlug(PageSource, JsonObject)` and `resolveName(PageSource)` so `RoqTemplateExtension` can call them directly without the `toLinkData` adapter.
3. **For `slugify` options**: expose `slugify(boolean preserveCase, boolean preservePath)` to match the existing `StringPaths` API, instead of the string-based `'lower'`/`'upper'` convention.

roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/utils/TemplateLink.java
ia3andy: @matheusandre1 we should add slugify to resolveName and remove it from here

ia3andy (changes requested):

roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/utils/TemplateLink.java
```diff
@@ -80,36 +80,43 @@ private static Map<String, Supplier<String>> withBasePlaceHolders(LinkData data,
...
+    private static String resolveName(LinkData data) {
+        return resolveName(data.pageSource());
+    }
+
+    public static String resolveSlug(PageSource pageSource, JsonObject data) {
+        String title = data.getString(SLUG,
+                data.getString(TITLE));
         if (title == null || title.isBlank()) {
-            title = resolveName(data);
+            title = resolveName(pageSource);
```
ia3andy:
```suggestion
            return resolveName(pageSource);
```

ia3andy (approved):
Thanks for you patience @matheusandre1. Nice work :)

PR #953, opened 2026-06-03
Use new junit dependency instead of old junit5 dependency

holly-cummins:
Should fix some warnings. If we're lucky, it might even speed up the build a touch.

### DIFF
--- a/blog/pom.xml
+++ b/blog/pom.xml
@@ -96,7 +96,7 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-junit5</artifactId>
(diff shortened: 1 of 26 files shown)


jtama (approved):
LGTM.
Thanks for the PR.

holly-cummins:
The CI failure is `2026-06-03T12:31:08.2609450Z [ERROR]   GitSyncServiceTest.tearDown:75->cleanDirectory:84 » UncheckedIO java.nio.file.NoSuchFileException: /tmp/roq-git-local-18076862536303305376/.git/gc.log.lock`

I think it's safe to assume this is unrelated.

Issue #366, opened 2025-01-13
Old site converter to Roq


ia3andy:
Experiment using AI to convert old site (such as Jekyll) to Roq

PR #963, opened 2026-06-05
Add Liquid-to-Qute template converter as first step of support for converting Jekyll sites to Roq

holly-cummins:
This is large, sorry. :( 

It's a standalone converter tool that transforms Jekyll/Liquid templates to Roq/Qute templates. Handles variables, filters, conditionals, loops, includes, assignments, case statements, layout tags, and special blocks.

I haven't hooked this into the docs yet, or into the larger migration logic, just to keep changes at a manageable size. 

### Limitations 
- Converting docs about Qute itself is hard, and probably doesn't work yet 
- There are two TODOs where the converter gives up and emits a TODO; I'll need to iterate on those, since they should be handled automatically, and even if they're not, I don't think a TODO in the output file is the right behaviour

### Good things 

It works on quarkus.io! "Works" is a strong word here: I didn't attempt the Qute docs, and I don't have the site fully up yet. 

Since Liquid is used more widely than just Jekyll, this converter may have a broader usefulness (cc @mkouba), but I'll store it here for now. 

I did look at conversion frameworks, like OpenRewrite, but for this kind of conversion I couldn't find anything that was obviously better than shell script + Java code. 

Partial resolution of #366

### DIFF
--- a/migration/src/main/java/io/quarkus/tools/LiquidToQuteCommand.java
+++ b/migration/src/main/java/io/quarkus/tools/LiquidToQuteCommand.java
@@ -31,7 +31,14 @@ public class LiquidToQuteCommand implements Callable<Integer> {
     @Option(names = {"-v", "--verbose"}, description = "Verbose output")
     private boolean verbose;
 
-    private final LiquidToQuteConverter converter = new LiquidToQuteConverter();
+    @Option(names = {"-e", "--extensions"}, description = "Template file extensions to process (default: .html, .htm, .liquid, .md, .markdown)", split = ",")
+    private List<String> templateExtensions = List.of(".html", ".htm", ".liquid", ".md", ".markdown");
+
+    @Option(names = {"--extension-syntax"}, description = "Use Qute extension syntax {=expr} instead of standard {expr} (default: true)",
+            defaultValue = "true", negatable = true)
+    private boolean extensionSyntax = true;
+
+    private LiquidToQuteConverter converter = new LiquidToQuteConverter();
 
     public static void main(String... args) {
         int exitCode = new CommandLine(new LiquidToQuteCommand()).execute(args);
@@ -40,6 +47,8 @@ public class LiquidToQuteCommand implements Callable<Integer> {
 
     @Override
     public Integer call() throws Exception {
+        converter = new LiquidToQuteConverter(extensionSyntax);
+
         if (!Files.exists(input)) {
             System.err.println("Error: Input path '" + input + "' does not exist");
             return 1;
@@ -101,8 +110,6 @@ public class LiquidToQuteCommand implements Callable<Integer> {
     }
 
     private void convertDirectory(Path inputDir, Path outputDir) throws IOException {
-        List<String> templateExtensions = List.of(".html", ".htm", ".liquid", ".md", ".markdown");
-
         int convertedCount = 0;
         int errorCount = 0;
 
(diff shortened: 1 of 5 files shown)


ia3andy:
Should we add it in jbang-ctalog?

ia3andy:
https://github.com/quarkiverse/quarkus-roq/blob/main/jbang-catalog.json

ia3andy:
in case doc for this is here: https://iamroq.dev/docs/migrating/

holly-cummins:
> Should we add it in jbang-ctalog?

I actually ended up going away from jbang on the next iteration of this (will be a follow-on PR), just because it needs a lot of unit tests and then having a fully-fledged project makes that easier. Because so much context and supporting scripting is needed I'm not sure many people would choose to run this from a jbang catalog, so I think we don't need to rush to publish it there. 

> in case doc for this is here: https://iamroq.dev/docs/migrating/

Yep! Updating and extending that is one of the other tasks on my todo list.

jtama:
> > Should we add it in jbang-ctalog?
> 
> I actually ended up going away from jbang on the next iteration of this (will be a follow-on PR), just because it needs a lot of unit tests and then having a fully-fledged project makes that easier. Because so much context and supporting scripting is needed I'm not sure many people would choose to run this from a jbang catalog, so I think we don't need to rush to publish it there.
> 

We can add it to the jbang catalog even if it's not just a script =>

```json
{
  "aliases": {
    "liquid-to-qute": {
      "script-ref": "https://repo1.maven.org/maven2/io/quarkiverse/roq/quarkus-roq-liquid-to-qute/2.1.3/quarkus-roq-liquid-to-qute-2.1.3-runner.jar",
      "description": "Migrate liqui tags to Qute ones",
      "java": "21+",
      "runtime-options": [
        "--add-opens",
        "java.base/java.lang=ALL-UNNAMED"
      ]
    }
  }
}
```

migration/src/main/java/io/quarkus/tools/LiquidToQuteCommand.java
```diff
@@ -0,0 +1,134 @@
...
+
+            return true;
+        } catch (IOException e) {
+            System.err.println("✗ Error converting " + inputPath + ": " + e.getMessage());
+            return false;
+        }
+    }
+
+    private void convertDirectory(Path inputDir, Path outputDir) throws IOException {
+        List<String> templateExtensions = List.of(".html", ".htm", ".liquid", ".md", ".markdown");
```
jtama: Shouldn't that be extensible ? I guess developer never stop being creatives.

jtama (commented):
ok I don't know jekyll, so nothing to do about this.

But in all replacement, your hypothesis is that the user uses Qute alternate syntax (`{=`), which could not be true. I think it should be a parameter with maybe default to true.

holly-cummins:
> ok I don't know jekyll, so nothing to do about this.
> 
> But in all replacement, your hypothesis is that the user uses Qute alternate syntax (`{=`), which could not be true. I think it should be a parameter with maybe default to true.

Ah, yes, that's a fair point. For quarkus.io, I think there's no chance of doing the conversion without the alternative syntax, and I have a wrapper that creates a config file which sets it. And I saw the default was changing, so just went with the alternative syntax. But you're totally right, we have the power of computers, so it should be configurable.

migration/src/main/java/io/quarkus/tools/LiquidToQuteCommand.java
holly-cummins: Good point. I'll add an argument for it.

holly-cummins:
I've added both of those configuration options. Good suggestions, thanks, @jtama. :)

jtama (approved):
LGTM

ia3andy (approved):
Thanks @holly-cummins :)

PR #983, opened 2026-06-16
Add tips on sanisation for Jekyll migration

holly-cummins:
As part of https://github.com/orgs/quarkusio/projects/79 and #366 I expect I'll have several updates to @rolfedh's excellent migration document. Here's a first one. 

I wasn't 100% sure on the structure, because there may be a range of different sanitisations people need to do and it may get too long for where it is in the document now.


ia3andy:
Should I merge, it looks good

rolfedh:
Very nice improvement. Thanks @holly-cummins!

jtama (approved):

PR #984, opened 2026-06-16
Expand on how to use bundled sass output

holly-cummins:
I had a bit of a hard time figuring out how to use sass with Roq. The page that explains it, https://iamroq.dev/web/sass/, wasn't turning up on a lot of my searches, for some reason. But once I did find it, it seemed to be missing a step about using the generated css. 

I've added what I *think* are the correct steps, but this is just based on experimentation, so I might be missing some nuance.

### DIFF
--- a/blog/content/marketplace/web/sass.md
+++ b/blog/content/marketplace/web/sass.md
@@ -34,8 +34,9 @@ $radius: 0.5rem;
 ```
 |}
 
-You can then reference the bundled stylesheet as `/static/bundle/app.css`, for example
-
+You can then reference the bundled stylesheet (i.e `/static/bundle/app-xxxxx.css`), via the bundle user tag (already included in base and other themes):
 ```html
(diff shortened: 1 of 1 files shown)


blog/content/marketplace/web/sass.md
```diff
@@ -32,4 +32,10 @@ $radius: 0.5rem;
...
 ```
-|}
\ No newline at end of file
+|}
+
+You can then reference the bundled stylesheet as `/static/bundle/app.css`, for example
+
+```html
+<link rel="stylesheet" href="/static/bundle/app.css" />
+```
```
ia3andy:
````suggestion
You can then reference the bundled stylesheet (i.e `/static/bundle/app-xxxxx.css`), via the bundle user tag (already included in base and other themes):
```html
{#bundle /}
```

See [the related documentation](/docs/basics/#styles-and-javascript).
````

jtama (approved):

PR #990, opened 2026-06-18
improved navigation

### LINKED ISSUE #989: back button in roq editor brings me back to extendsions list instead of posts
When press the back button after navigating from the list to a specific post, I end up on the home screen instead of back to the list of posts

edewit:
fixes: #989
Signed-off-by: Erik Jan de Wit <erikjan.dewit@gmail.com>

### DIFF
--- a/roq-editor/deployment/src/main/resources/dev-ui/qwc-roq-editor.js
+++ b/roq-editor/deployment/src/main/resources/dev-ui/qwc-roq-editor.js
@@ -73,6 +73,8 @@ export class QwcRoqEditor extends LitElement {
         this._syncing = false;
         this._publishing = false;
         this._syncManager = null;
+        this._fromHistoryNavigation = false;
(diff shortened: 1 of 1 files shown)


jtama:
Don't have anything to say on the code itself, the commit description might be a bit more self explaining ;)

jtama (changes requested):
@edewit could you please amend you commit message to be a bit more self explaining ? I had to read the whole code to understand what this is about.

edewit:
@jtama I can update the commit message, but I'm wondering why that would be needed as this is linked to an issue that describes what is wrong and this commit fixes that issue, no?

jtama:
> @jtama I can update the commit message, but I'm wondering why that would be needed as this is linked to an issue that describes what is wrong and this commit fixes that issue, no?

Thanks I do think it make things easier, for automation, the first one beeing the release note.

jtama (approved):

PR #991, opened 2026-06-18
Jekyll migration: Add converter to transform _config.yml 

holly-cummins:
This converter handles taking the Jekyll `_config.yml` and converting it into application.properties (for values which would be Roq config) and the `data/siteConfig.yml` used by the liquid converter.

### DIFF
--- a/migration/src/main/java/io/quarkus/tools/JekyllConfigConverter.java
+++ b/migration/src/main/java/io/quarkus/tools/JekyllConfigConverter.java
@@ -1,0 +1,166 @@
+package io.quarkus.tools;
+
+import java.io.IOException;
+import java.io.Writer;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Properties;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
+
+import static io.quarkus.tools.LiquidToQuteConverter.DATA_NAME;
+
+/**
+ * Converts Jekyll _config.yml to Roq application.properties and data/siteConfig.yml.
+ * Replaces the bash script logic from roq-it-jekyll lines 61-62, 184-192, and 238-277.
+ */
+public class JekyllConfigConverter {
+
+    private final YAMLMapper yamlMapper;
+    private final ObjectMapper objectMapper;
+
+    public JekyllConfigConverter() {
+        this.yamlMapper = new YAMLMapper();
+        this.objectMapper = new ObjectMapper();
+    }
+
+    /**
+     * Create application.properties values with standard Roq properties for Jekyll compatibility.
+     * Replaces roq-it-jekyll lines 184-192.
+     *
+     * @return Application properties
+     */
+    public Properties createApplicationProperties() {
+        Properties properties = new Properties();
+
+        // Always enable the alternative expression syntax to reduce overhead of escaping braces
+        properties.setProperty("quarkus.qute.alt-expr-syntax", "true");
+        // Set a date format with a sensible default for Jekyll.
+        properties.setProperty("site.date-format", "yyyy-MM-dd['T'HH:mm:ss][X]");
+        // Jekyll templates access arbitrary frontmatter fields that may not exist on every page
+        // echo "quarkus.qute.strict-rendering=false" >> ${project_dir}/config/application.properties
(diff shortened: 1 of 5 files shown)


migration/src/main/java/io/quarkus/tools/JekyllConfigConverter.java
```diff
@@ -0,0 +1,166 @@
...
+        siteConfig.put("cname", cnameContent != null ? cnameContent.trim() : "");
+        
+        // Copy common properties
+        copyIfPresent(config, siteConfig, "baseurl");
+        copyIfPresent(config, siteConfig, "language");
+        copyIfPresent(config, siteConfig, "twitter_username");
+        copyIfPresent(config, siteConfig, "github_username");
+        
+        // Handle nested search config
+        if (config.has("search")) {
```
jtama: Not really sure what this is for.

Roq has a lunr search plugin that doesn't use any of those properties. Is this specific to jekyll ? How is it gonna be intergrated to roq ?

migration/src/main/java/io/quarkus/tools/JekyllConfigConverter.java
```diff
@@ -0,0 +1,166 @@
...
+        // Build a new config with selected properties
+        Map<String, Object> siteConfig = new LinkedHashMap<>();
+        
+        // Add CNAME
+        siteConfig.put("cname", cnameContent != null ? cnameContent.trim() : "");
+        
+        // Copy common properties
+        copyIfPresent(config, siteConfig, "baseurl");
+        copyIfPresent(config, siteConfig, "language");
+        copyIfPresent(config, siteConfig, "twitter_username");
```
jtama: Usually, those are either in the index.html frontmatter tags (https://github.com/quarkiverse/quarkus-roq/blob/6e954631a7a3bf487e0d66d84354ad67c738bf4d/blog/content/index.html#L1) or in the authors.yaml file : (https://github.com/quarkiverse/quarkus-roq/blob/main/blog/data/authors.yml).

The question is : do we migrate existing sites as-is, or do we try to lead authors to the roq way of doing things ? Not sure I have a right answer though.

holly-cummins: > The question is : do we migrate existing sites as-is, or do we try to lead authors to the roq way of doing things ? Not sure I have a right answer though.

I've wondered the same thing, for other aspects of the conversion. Like, Roq has config flags for directory locations, so it could be configured to use the existing Jekyll file layout, which would mean a smaller diff after migration. But on that one, I decided people don't migrate to Roq because they want to continue using Jekyll, so I moved the file layouts. 

In this particular context, I'm not sure `authors.yaml` makes sense, because they're properties of the blog as a whole. But index.html makes sense, and I don't think there's too much dilemma about being idiomatic, because I'm moving them anyway. My one question is whether they'll be available to the places that need them if I put them in the index.html frontmatter, but I will try it and find out!

migration/src/main/java/io/quarkus/tools/JekyllConfigConverter.java
holly-cummins: This is actually specific to quarkus.io, which probably isn't better. :)

I've updated it locally to not hardcode the specific property names and just copy everything with a `search` prefix. But I should perhaps generalise it further. I've tried to isolate quarkus.io things into the `haq-it` script and not let them leak into the general migration scripts, but there's a general "your site-specific stuff moves from _config.yml to other places" principle that wants to be applied.

migration/src/main/java/io/quarkus/tools/JekyllConfigConverter.java
jtama: The availability is a good point. They _should_ be in `site.data`

holly-cummins:
Hi @jtama, thanks for the thoughtful review. I'm adding a bunch of changes, which sort of address your comments, but also do a lot more (doh!). I've gone fully generic, rather than special-casing various keys and `search`.  The problem the search transform was solving was actually generic, and could affect other keys. 

The problem is qute interprets hyphens as the subtraction operator. In Jekyll, configs use hyphens (yaml convention):
```
  search:
    script-mode: "cached"
    initial-timeout: 1500
  ```

In Qute templates, you would access these as:

```
  {cdi:siteConfig.search.script-mode} 
```

Boom splat. Qute interprets this as: search.script minus mode, which is nonsense. I can't think of a downside of getting rid of all hyphenated keys (below the top level) so I've done that.

ia3andy:
@holly-cummins can merge?

holly-cummins:
> @holly-cummins can merge?

Yes please!

PR #979, opened 2026-06-12
Add script to convert Jekyll sites to Roq

holly-cummins:
Resolves #366 . 

For this to run cleanly, various other PRs need merging, but the script itself is complete. 

There are, in my mental model, four parts to a conversion (https://iamroq.dev/docs/migrating/ lists four, but I dropped one and added one from that list):
- Setting up project scaffolding (done in the shell script)
- Adjusting layout from Jeykll style to Roq style (done in the shell script)
- Converting templates from Liquid to Qute (done in Java, see #963, but this is the biggest part, so more needed)
- Extracting site config and moving it to Roq/Quarkus format (done a bit this shell script, but more to do)

I'm also planning to add a `roq-it` script which does detection of the source type, so people can run that directly. It would be nice to eventually have converters for Mkdocs (#1118) and Hugo (#1114 ) as well as Jekyll. 

I'd like to move away from shell script orchestration to a Java orchestrator, and eventually to a `roq` cli extension so people can do `roq migrate`, but I don't want to do that kind of refactoring until there's something working in main with regular CI runs (#1006). 

Obviously, this all needs docs updates, but I will do those once everything is in main.

PR #998, opened 2026-06-19
Add utilities for converting Jekyll frontmatter to Roq equivalents

holly-cummins:
To get permalinks converted to aliases, and to convert pagination instructions from Jekyll format to Roq format, we need some conversion utilities.

### DIFF
--- a/migration/src/main/java/io/quarkus/tools/migration/jekyll/JekyllFrontMatterCommand.java
+++ b/migration/src/main/java/io/quarkus/tools/migration/jekyll/JekyllFrontMatterCommand.java
@@ -1,0 +1,51 @@
+///usr/bin/env jbang "$0" "$@" ; exit $?
+//DEPS info.picocli:picocli:4.7.5
+//DEPS com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.18.2
+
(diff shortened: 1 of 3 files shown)


jtama:
I will merge but shouldn't we have a super command like: `migrate jekyll` ?

holly-cummins:
> I will merge but shouldn't we have a super command like: `migrate jekyll` ?

Definitely. Well, something of that kind. I'd been assuming a shell script for the directory manipulation and general orchestration, but I see your point, maybe everything should be a Java command.

What I've got so far is `roq-it-jekyll` (there's a draft in https://github.com/quarkiverse/quarkus-roq/pull/979), and I was assuming once it was in a good state, I'd add a `roq-it` command which would look to see what kind of site it was (Jekyll, Gatsby, etc) and call the right sub-script. I also have, for my own personal use, `haq-it`, which wraps `roq-it` but does a bunch of git resetting and deletes files that don't convert yet. The name of that makes me smile every time. :) 

But thinking about it, `roq`as a verb is pretty overloaded in the Roq context (we have `@RoqAndRoll` as well), so `migrate` would be less entertaining, but more clear. I'll make a note over on #979.

jtama:
`haq-it` is fine for me ;)

PR #1003, opened 2026-06-24
Evaluate (sort of) conditional directives in asciidoc before yupiik header parsing

holly-cummins:
I noticed that even if I use the asciidoc-jruby plugin, ifdefs in asciidoc headers cause crashes. 

```
  java.lang.IllegalArgumentException: Unknown line: 'ifdef::env-github,env-browser,env-vscode[:imagesdir: ../assets/images/posts]'
      at io.yupiik.asciidoc.parser.Parser.readAttributes(Parser.java:1634)
      at io.yupiik.asciidoc.parser.Parser.parseHeader(Parser.java:209)
```

It turns out that the `AsciidocHeaderParser` in `asciidoc-common` uses yupiik under the covers, and yupiik can't handle conditionals. How asciidoc handles this is to strip out the conditionals before doing the parsing. I've mirrored this approach, in a slightly simplified way. Most `env-` attributes would always be false inside a Roq environment (it's not github, or vscode, or ...), so I've just defaulted those to false. Attributes passed in via build configuration *could* be true, but we don't have access to them so I've also cheated and defaulted them to false. I did add some code which reads the rest of the header and honours any attributes defined there, because it would look silly not to. 

My first implementation used several regexes. I felt a bit uneasy about doing so much parsing just with regular expressions, but I checked, and asciidoc also uses good old-fashioned regexes for its parsing. They have a single regex, which makes maintenance easier, if not readability. I borrowed the asciidoc regex on the basis that it must be correct. :) 

We didn't seem to have tests for asciidoc-common, so I've added the dependencies and tests for these cases in the header parser.

### DIFF
--- a/roq-plugin/asciidoc-common/deployment/src/main/java/io/quarkiverse/roq/plugin/asciidoc/common/deployment/AsciidocHeaderParser.java
+++ b/roq-plugin/asciidoc-common/deployment/src/main/java/io/quarkiverse/roq/plugin/asciidoc/common/deployment/AsciidocHeaderParser.java
@@ -57,9 +57,9 @@ public class AsciidocHeaderParser {
     public static RoqFrontMatterHeaderParserBuildItem createBuildItem(boolean qute, Predicate<TemplateContext> isApplicable) {
         return new RoqFrontMatterHeaderParserBuildItem(isApplicable, templateContext -> {
             Parser parser = new Parser();
-            String content = templateContext.content();
+            String content = stripFrontMatter(templateContext.content());
             if (SIMPLE_CONDITIONAL.matcher(content).find()) {
-                content = stripConditionalDirectives(stripFrontMatter(content));
+                content = stripConditionalDirectives(content);
             }
 
             ContentResolver contentResolver = new PathContentResolver(templateContext.sourceFile().getParent());
--- a/roq-plugin/asciidoc-common/deployment/src/main/java/io/quarkiverse/roq/plugin/asciidoc/common/deployment/AsciidocHeaderParser.java
+++ b/roq-plugin/asciidoc-common/deployment/src/main/java/io/quarkiverse/roq/plugin/asciidoc/common/deployment/AsciidocHeaderParser.java
@@ -9,12 +9,19 @@ import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Deque;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.jboss.logging.Logger;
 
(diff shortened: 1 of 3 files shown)


ia3andy:
I am a bit worried about this bits of code because of the perf cost of doing this for every asciidoc files, it could also break the content. Could we somehow mitigate this with a if check before to apply it only when some keywords are present in the header?

holly-cummins:
> I am a bit worried about this bits of code because of the perf cost of doing this for every asciidoc files, it could also break the content. Could we somehow mitigate this with a if check before to apply it only when some keywords are present in the header?

That's a good idea. The marker is `ifdef` and friends - basically, if there's an `ifdef` and we don't do something, Roq crashes. So we could only do the header ingestion when there's an ifdef,

Another option is to say that it's relatively rare that ifdef would be guarded on another property in the header attributes, and we just don't attempt to resolve it at all (and always return false). That's what we're already doing for environment conditions (like `env-github`) and properties passed in. So it would have the advantage of less code, and a slightly more consistent behaviour – at the expense of doing the wrong thing in a case when we _could_ do the right thing. But that case might be rare.

jtama:
Using ifdef on something defined in the header document is kind of a non-sens, a bit like `if true`...

holly-cummins:
> Using ifdef on something defined in the header document is kind of a non-sens, a bit like `if true`...

Yeah, it's hard to see the value of it. I guess it's useful for copy-paste scenarios, like you just repeat a header but then you want to nudge behaviour in one file in an explicit way so you do an override in a header. But I agree it's probably not worth taking a performance hit to support it.

holly-cummins:
Actually, I changed my mind again. :) So I've put in the short-circuit suggested by @ia3andy, but kept the attribute evaluation. Some of the examples in the tests looked a bit too much like asciidoc I've seen, and I thought it would confuse people if we did the wrong things in those cases. 

We could profile and take it out if it's a performance drag.

roq-plugin/asciidoc-common/deployment/src/main/java/io/quarkiverse/roq/plugin/asciidoc/common/deployment/AsciidocHeaderParser.java
```diff
@@ -30,13 +37,31 @@
...
+
+    // Matches attribute entries: :name: value, :!name: (unset), :name!: (unset)
+    private static final Pattern ATTRIBUTE_ENTRY = Pattern.compile("^:(!?\\w[^:]*):(?:[ \\t]+(.*))?$");
+
     public static RoqFrontMatterHeaderParserBuildItem createBuildItem(boolean qute, Predicate<TemplateContext> isApplicable) {
         return new RoqFrontMatterHeaderParserBuildItem(isApplicable, templateContext -> {
             Parser parser = new Parser();
-            String content = stripFrontMatter(templateContext.content());
+            String content = templateContext.content();
+            if (SIMPLE_CONDITIONAL.matcher(content).find()) {
```
ia3andy, line 61: I think you removed `stripFrontMatter(content)` by mistake?

ia3andy (changes requested):

roq-plugin/asciidoc-common/deployment/src/main/java/io/quarkiverse/roq/plugin/asciidoc/common/deployment/AsciidocHeaderParser.java
holly-cummins: Ooh dear, it does look like it, doesn't it? One too many edits, it seems – but why are the tests still passing? :(

holly-cummins: Added the missing test (which fails on what you reviewed), and added back `stripFrontMatter`. Thanks!

ia3andy (approved):
LGTM

PR #1004, opened 2026-06-24
Hook migration into top-level build and parent

holly-cummins:
Part of #366. 

The migration module wasn't hooked into the top level build as a module. This is maybe sensible when I first created it, but it's now got enough stuff in it we should start running it as part of the builds. It also didn't use the roq parent, so I've fixed that. 

A side-effect of adding the parent was a whole bunch of formatting churn, but hopefully it's a one-off. 

Another side-effect is that I noticed one test is failing, and CI didn't catch it, because the migration module wasn't being CI-ed. I've got a bunch of changes on another branch for the converter, so in this branch, I just disabled the failing test.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,6 +25,7 @@
         <module>roq-plugin</module>
         <module>roq-theme</module>
         <module>roq-cli</module>
+        <module>migration</module>
     </modules>
 
     <scm>
@@ -130,7 +131,6 @@
             </activation>
             <modules>
                 <module>blog</module>
-                <module>migration</module>
             </modules>
         </profile>
     </profiles>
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -130,6 +130,7 @@
             </activation>
             <modules>
                 <module>blog</module>
+                <module>migration</module>
```
jtama: So that means that the migration tool isn't released. Is it on purpose ?

holly-cummins: I see what you mean. No, probably not! It's not ready for release yet, but it should be in that section of the pom. Will fix, thanks!

jtama (approved):

PR #1006, opened 2026-06-24
Add migration e2e tests

holly-cummins:
I'd like to have end-to-end testing of the conversion, so taking a Jekyll site, converting, and then making sure it's a functional Roq site. I initially wondered about running `jekyll new` but that doesn't come with any templates, so I've created what I hope is an small idiomatic site from scratch. That does mean the PR is bigger and we have to source control a bunch of Jekyll code. 

The tests can be run using `@RoqAndRoll` or against an external site, which allows us to confirm the same tests pass on both Jekyll and Roq sites.

While getting this working, I discovered a few problems:

- The script prereqs jbang, so I added an install step
- The script just doesn't work on windows, at all. I tried to play with slashes but eventually gave up; we want to move away from a shell script, and this is another good reason. So I disabled the tests on windows, for now.
- Some steps invoked by the script are in outstanding PRs. I added guards so this can merge without being blocked.
- The script runs tests while executing, which is silly. Well, it's silly until a test runs the script, and then it's an infinite loop which caused my laptop to consume 48 pages of swap, lock up, and reboot itself every time I tried to run the tests. 

Here are the commands to run against jekyll:

#### 1. Start Jekyll on the fixture
```
   cd migration/src/test/resources/jekyll-site
   bundle install
   bundle exec jekyll serve --port 4000
```

#### 2. In another terminal, run E2E tests against Jekyll
```
     cd migration
    mvn failsafe:integration-test -Dtest.base.url=http://localhost:4000
```


holly-cummins:
Do you think you'll have time to review this, @ia3andy or @jtama? I'd like to get the tests merged to provide safety while I switch the converter from scripts to Java.

jtama:
Sorry not before a few days.

migration/src/test/java/io/quarkus/tools/JekyllConversionTestResource.java
```diff
@@ -0,0 +1,143 @@
...
+            System.err.println("[TEST-RESOURCE] workDir exists = " + Files.exists(absWorkDir));
+            System.err.println("[TEST-RESOURCE] content dir exists = " + Files.exists(absWorkDir.resolve("content")));
+            System.err.println("[TEST-RESOURCE] templates dir exists = " + Files.exists(absWorkDir.resolve("templates")));
+            if (Files.exists(absWorkDir.resolve("content"))) {
+                try (var walk = Files.walk(absWorkDir.resolve("content"), 3)) {
+                    walk.forEach(p -> System.err.println("[TEST-RESOURCE] content: " + absWorkDir.relativize(p)));
+                }
+            }
+
+            return Map.of("quarkus.roq.dir", absWorkDir.toString());
```
jtama, line 66: Not sure I understand why a Map si returned and not a string...

migration/src/test/java/io/quarkus/tools/JekyllToRoqConversionTest.java
```diff
@@ -0,0 +1,209 @@
...
+        assertThat(workDir.resolve("content/posts")).isDirectory();
+        assertThat(workDir.resolve("templates/layouts")).isDirectory();
+        assertThat(workDir.resolve("templates/partials")).isDirectory();
+        assertThat(workDir.resolve("web/main.css")).exists();
+        assertThat(workDir.resolve("config/application.properties")).exists();
+
+        // --- Verify config ---
+
+        String config = Files.readString(workDir.resolve("config/application.properties"));
+        assertThat(config)
```
jtama, line 136: Seems to me there are WAYYYYYY to much assert for just one test case, but I'm reading diagonally...

migration/src/test/java/io/quarkus/tools/JekyllToRoqConversionTest.java
```diff
@@ -0,0 +1,209 @@
...
+                .contains("quarkus.qute.strict-rendering=false");
+    }
+
+    private void copyFixture(Path target) throws IOException {
```
jtama, line 140: Already exists is JekyllConversionTestResource. If I'm not wrong, true for every private method.

migration/src/test/resources/application.properties
```diff
@@ -0,0 +1,5 @@
+quarkus.roq.dir=target/converted-jekyll-site
+quarkus.qute.alt-expr-syntax=true
+quarkus.qute.strict-rendering=false
+quarkus.qute.type-check-excludes=java.lang.Object.*,io.quarkiverse.roq.frontmatter.runtime.model.DocumentPage.*,io.quarkiverse.roq.frontmatter.runtime.model.Page.*,io.quarkiverse.roq.frontmatter.runtime.model.Page.paginator,io.quarkiverse.roq.frontmatter.runtime.model.Page.tags,io.quarkiverse.roq.frontmatter.runtime.model.Page.tagsCount
```
jtama, line 4: Never used that? Is it really necessary ?

jtama (changes requested):
I've made a few comments, not sure haow critical they are though.

migration/src/test/resources/application.properties
holly-cummins: The converted quarkus.io has it. I forget exactly why it's there, but I just tried removing it, and now I see :) 

```
                ... 10 more                                                                                                                                                                                                                
        Suppressed: io.quarkus.qute.TemplateException: partials/blog-band.html:26:59 - {author.name.raw}: Property/method [name] not found on class [java.lang.Object] nor handled by an extension method                                  
                at io.quarkus.qute.TemplateException$Builder.build(TemplateException.java:173)                                                                                                                                             
                at io.quarkus.qute.deployment.QuteProcessor.processTemplateErrors(QuteProcessor.java:265)                                                                                                                                  
                ... 10 more         
```

I guess in an ideal world maybe there would be more type safety/assignment somewhere, which would make the exclusions unnecessary. Qute is treating everything in the html templates as Objects, and then getting unhappy when template extensions are invoked.

migration/src/test/java/io/quarkus/tools/JekyllMigrationE2EIT.java
```diff
@@ -0,0 +1,129 @@
...
+ *   # Against a manually started Roq site:
+ *   cp -r migration/src/test/resources/jekyll-site /tmp/my-roq-site
+ *   ./migration/roq-it-jekyll /tmp/my-roq-site
+ *   mvn -f migration verify -Dtest.base.url=http://localhost:8080
+ * }</pre>
+ */
+@QuarkusTest
+@RoqAndRoll
+@DisabledOnOs(value = OS.WINDOWS, disabledReason = "Depends on JekyllConversionTestResource which requires bash")
+class JekyllMigrationE2EIT {
```
ia3andy, line 67: have you tried the playwright extension instead?

ia3andy:
Two things that stand out:

**Fixture file list**: `FIXTURE_FILES` is hardcoded (and duplicated between `JekyllConversionTestResource` and `JekyllToRoqConversionTest`). If someone adds a file to the Jekyll fixture, both lists silently go stale. Could you just walk the fixture directory at copy time instead?

**`BashCommandHelper`**: This re-implements bash discovery and path conversion from scratch. Quarkus core already has `io.quarkus.utilities.OS` and similar utilities for this kind of platform detection. Even though Windows is disabled for now, if it gets enabled later, I'd rather rely on existing tested code than maintain a custom helper. Could you check what's available in Quarkus core before rolling your own?

Also, have you tried `quarkus-playwright` (`@WithPlaywright`) instead of managing the Playwright lifecycle manually? It handles browser setup/teardown for you and is already part of the Quarkiverse ecosystem.

migration/src/test/java/io/quarkus/tools/JekyllConversionTestResource.java
holly-cummins: This is just what the parent interface, `QuarkusTestResourceLifecycleManager.start()` takes.

migration/src/test/java/io/quarkus/tools/JekyllToRoqConversionTest.java
holly-cummins: I agree, I've split it up. I prefer smaller test cases but sometimes there's a tradeoff against either redoing work and having slowness, or moving work into `BeforeEach` that should really be in a test, or adding a required execution order. Here I don't think any of those apply, so I've made it smaller, as it should have been to begin with.

migration/src/test/java/io/quarkus/tools/JekyllToRoqConversionTest.java
holly-cummins: Good spot, I've combined.

migration/src/test/java/io/quarkus/tools/JekyllMigrationE2EIT.java
holly-cummins: I ruled it out for this, because the tests have to be able to run against both Quarkus and Jekyll. I know there's the `RoqAndRoll` and `QuarkusTest` annotations which are Quarkus-y, but it does work to run them on the CLI and pass in the URL. I think using the Playwright extension would make the code smaller, but interfere with that ability.

holly-cummins:
> **Fixture file list**: `FIXTURE_FILES` is hardcoded (and duplicated between `JekyllConversionTestResource` and `JekyllToRoqConversionTest`). If someone adds a file to the Jekyll fixture, both lists silently go stale. Could you just walk the fixture directory at copy time instead?

Good point, done!

> **`BashCommandHelper`**: This re-implements bash discovery and path conversion from scratch. Quarkus core already has `io.quarkus.utilities.OS` and similar utilities for this kind of platform detection. Even though Windows is disabled for now, if it gets enabled later, I'd rather rely on existing tested code than maintain a custom helper. Could you check what's available in Quarkus core before rolling your own?

For this, I think we don't want to bother, because I'm hoping to get rid of the shell scripts soon. I've checked the before and after, and it's fairly marginal (probably not worth the extra dependency). `BashCommandHelper` doesn't go away, it's just the OS-detection that becomes nicer:

Before:
```
static String getBashCommand() {
    String os = System.getProperty("os.name").toLowerCase();
    if (os.contains("win")) {
        String[] possiblePaths = { ... };
        for (String path : possiblePaths) {
            if (Files.exists(Path.of(path))) return path;
        }
        return "bash";
    }
    return "bash";
}
```

After:
```
static String getBashCommand() {
    if (OS.WINDOWS.isCurrent()) {
        String[] possiblePaths = { ... };
        for (String path : possiblePaths) {
            if (Files.exists(Path.of(path))) return path;
        }
        return "bash";
    }
    return "bash";
}
```
 
> Also, have you tried `quarkus-playwright` (`@WithPlaywright`) instead of managing the Playwright lifecycle manually? It handles browser setup/teardown for you and is already part of the Quarkiverse ecosystem.

See discussion above - I think it couples the test to Quarkus too much.

holly-cummins:
Is this ready to go in? I'd like to finish off the converter work before I forget all context about what I'd done and why. :)

PR #1008, opened 2026-06-25
Add tests specifically checking slug overrides are honoured

holly-cummins:
We have test coverage user-overrides of slugs for collections (in `RoqFrontMatterRoutingTest#testHtmlPost`), but not for normal pages. Slugs aren't currently honoured for normal pages, so this adds the missing tests (which fail, because the slug isn't honoured). 

I've raised an issue to track this at https://github.com/quarkiverse/quarkus-roq/pull/1009.


holly-cummins:
Actually, I'll make the new tests disabled, and then we can merge this and un-disable them when we fix #1009.

ia3andy (approved):
thanks

PR #1011, opened 2026-06-26
Support deeply nested data directories as grouped CDI beans

holly-cummins:
This is needed for quarkus.io, but I think it will be more generally useful, too; it doesn't fix a todo, but it does fix a debug log. The quarkusio.github.io guides listing page uses data files with several levels of nesting, such as `data/versioned/latest/index/quarkus.yaml`. The template accesses this as cdi:versioned.get('latest').index.quarkus.

Roq's data processor only grouped single-level directories (e.g. `data/heroes/batman.yaml` → `cdi:heroes.batman`). Anything deeper was skipped with a debug log. So `cdi:versioned` was never produced as a CDI bean. This was bad. :)

With the fix, data files in nested subdirectories are now grouped into a single CDI bean named after the top-level directory, with the directory structure preserved as nested `JsonObject`s.

### DIFF
--- a/roq-data/deployment/src/main/java/io/quarkiverse/roq/data/deployment/RoqDataReaderProcessor.java
+++ b/roq-data/deployment/src/main/java/io/quarkiverse/roq/data/deployment/RoqDataReaderProcessor.java
@@ -185,19 +185,21 @@ public class RoqDataReaderProcessor {
             }
         }
 
-        // Produce grouped RoqDataJsonBuildItem for untyped directories
-        for (Map.Entry<String, TreeMap<String, RoqDataBuildItem>> entry : allDirFiles.entrySet()) {
-            if (!producedJsonNames.contains(entry.getKey()) && !dirAnnotationNames.contains(entry.getKey())) {
-                TreeMap<String, Object> grouped = new TreeMap<>();
-                for (Map.Entry<String, RoqDataBuildItem> fileEntry : entry.getValue().entrySet()) {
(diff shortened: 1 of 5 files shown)


roq-data/deployment/src/main/java/io/quarkiverse/roq/data/deployment/RoqDataReaderProcessor.java
```diff
@@ -349,6 +351,29 @@ private void processDirectoryAnnotations(
...
+                String dir = key.substring(0, slashIdx);
+                String rest = key.substring(slashIdx + 1);
+                subDirs.computeIfAbsent(dir, k -> new TreeMap<>()).put(rest, entry.getValue());
+            } else {
+                result.put(key, entry.getValue());
+            }
+        }
+
+        for (Map.Entry<String, TreeMap<String, Object>> subDir : subDirs.entrySet()) {
+            result.put(subDir.getKey(), buildNestedJsonObject(subDir.getValue()));
```
jtama, line 371: So we can nest directory indefinitely ?

jtama (approved):
LGTM

ia3andy (approved):
Cool!

roq-data/deployment/src/main/java/io/quarkiverse/roq/data/deployment/RoqDataReaderProcessor.java
holly-cummins: That's the idea ...I *think* the tests cover more than one level deep.

PR #1013, opened 2026-06-26
Add :dir placeholder to support slug overrides and nested paths

### LINKED ISSUE #1005: Paths in collections get flattened
This is another issue I spotted working on quarkus.io. 

It's fixable, but I didn't want to charge off and fix it without checking in on what approach we prefer. 

The problem happens because of a difference in the default URL template for collection documents:
     
  - Jekyll: uses the file path within the collection → /:collection/:path/ → /versions/2.13/guides/amqp-dev-services/ (preserves directory structure)
  - Roq: uses /:collection/:slug/ → /versions/dev-services-for-amqp/ (slug from title, flattens everything)

  Jekyll's defaults naturally keeps each version's guides unique. Roq's :slug comes from the title (or filename), which is the same across version directories, so they all collapse to the same URL.

It feels correct to make the default doc link template path-aware. But if we don't want to change the default, we definitely need to provide the right placeholders so that people can set the paths in a per-file way (although that could be a lot of repeated boilerplate). 

Looking at Roq's existing placeholders, `:path` already preserves directory structure but includes the collection prefix (so `/:collection/:path/` would double it to `/versions/versions/2-13/...`). There's no placeholder  for "path within the collection minus the collection prefix."

I think we could do one of two things, if we don't want to break the behaviour of existing placeholders:

  1. Add a `:collection-path` placeholder which would be the slugified document path relative to the collection directory (without extension), e.g. `2.13/guides/amqp-dev-services`.

  2. Strip the collection prefix from `:path` automatically when the document is in a collection and the template uses both `:collection` and `:path`.

Then, assuming option 1, I'd also vote we change 
```
    public static final String DEFAULT_PAGE_LINK_TEMPLATE = "/:path:ext";
    public static final String DEFAULT_DOC_LINK_TEMPLATE = "/:collection/:slug/";
```

to 

```
    public static final String DEFAULT_PAGE_LINK_TEMPLATE = "/:path:ext";
    public static final String DEFAULT_DOC_LINK_TEMPLATE = "/:collection-path:ext";
```

which is more consistent. That also fixes another inconsistency in using `:slug` in one case and not the other, but then it means user-defined slug attributes are never honoured, which is kind of weird. I don't think I understand why`DEFAULT_PAGE_LINK_TEMPLATE` doesn't use the slug, but I guess if it did, `:path` would have to stop including the file name. To me, that's most logical, but it would be a big breaking change. 
 
For option 2, if we changed the defaults, it would be 

```
    public static final String DEFAULT_PAGE_LINK_TEMPLATE = "/:path:ext";
    public static final String DEFAULT_DOC_LINK_TEMPLATE = "/:collection/:path:ext/";
```

... so there's still the oddness with the non-honoured slugs. 

We could also say "working as designed" and I can tactically set a `link:` on every page, but I don't love that, because it feels like when people create folder hierarchies, they expect them to be reflected in the published site.

holly-cummins:
Resolves #1005 

This adds a `:dir` placeholder which can be used to support non-collapsed collection paths. It is not used in the default, to continue to allow internal sorting patterns like `posts/2025` to work. 

The `:dir` placeholder strips `/draft/` segments, consistent with other placeholders. 

Ideally the path parsing would use utilities in `StringPath`, but the right utilities don't exist yet, so I didn't want to let that block this.

### DIFF
--- a/blog/content/docs/advanced.adoc
+++ b/blog/content/docs/advanced.adoc
@@ -250,6 +250,11 @@ TIP: Those links are also available in the Qute data to allow <<roq-url>>.
 | The raw file path of the page without the extension.
 | `My$`, `my car` or `été/2024`
 
+| All
+| `:dir`
+| The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root.
+| `posts`, `posts/v2/guides`, or empty
+
 
 | All
 | `:slug`
--- a/roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/util/RoqFrontMatterAssembleUtils.java
+++ b/roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/util/RoqFrontMatterAssembleUtils.java
@@ -1,7 +1,8 @@
 package io.quarkiverse.roq.frontmatter.deployment.util;
 
-import static io.quarkiverse.roq.frontmatter.deployment.util.RoqFrontMatterLayoutUtils.*;
-import static io.quarkiverse.roq.frontmatter.deployment.util.RoqFrontMatterTemplateUtils.*;
+import static io.quarkiverse.roq.frontmatter.deployment.util.RoqFrontMatterLayoutUtils.resolveDefaultLayout;
+import static io.quarkiverse.roq.frontmatter.deployment.util.RoqFrontMatterTemplateUtils.applyContentTransforms;
+import static io.quarkiverse.roq.frontmatter.deployment.util.RoqFrontMatterTemplateUtils.getMarkup;
 import static io.quarkiverse.roq.frontmatter.runtime.RoqFrontMatterKeys.ESCAPE;
 import static io.quarkiverse.roq.frontmatter.runtime.RoqFrontMatterKeys.LAYOUT;
 import static io.quarkiverse.roq.frontmatter.runtime.RoqFrontMatterKeys.THEME_LAYOUT;
--- a/roq-frontmatter/deployment/src/test/java/io/quarkiverse/roq/frontmatter/deployment/util/TemplateLinkTest.java
+++ b/roq-frontmatter/deployment/src/test/java/io/quarkiverse/roq/frontmatter/deployment/util/TemplateLinkTest.java
@@ -153,6 +153,7 @@ class TemplateLinkTest {
         assertDoesNotThrow(() -> pageLink("", "/:year/:month/:day/:slug", data));
         assertDoesNotThrow(() -> pageLink("", "/:path:ext", data));
         assertDoesNotThrow(() -> pageLink("", "/:name/:Slug", data));
+        assertDoesNotThrow(() -> pageLink("", "/:dir/:slug", data));
     }
 
     @Test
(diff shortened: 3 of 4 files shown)


ia3andy:
@holly-cummins we can't generate multiple routes, it would just either duplicate the generated html page which I don't think we want. We should be extra-careful about not breaking routing for existing sites.
I think we shouldn't honour directory structure by default, the current default is fine. Quarkus should define it's own custom link via layout or config.

holly-cummins:
> @holly-cummins we can't generate multiple routes, it would just either duplicate the generated html page which I don't think we want. We should be extra-careful about not breaking routing for existing sites. I think we shouldn't honour directory structure by default, the current default is fine. Quarkus should define it's own custom link via layout or config.

Most of the potential breaks come from the change to honour user-defined slugs, not the not-collapsing-directory-structure change. But I 100% agree we should be very careful to avoid breaking existing sites. That's exactly why I added the duplicate routes. They'd only duplicate a page in a case where the old route is different from the new one, so it should be rare. I agree the route duplication is kind of ugly, but I don't see a way to honour slugs and also honour the old never-use-the-slug routes in the case where the end up different, except by having two routes.

holly-cummins:
Note also that not collapsing urls (the current default) creates a lot of potential for name collisions. If a user creates 
`posts/path1/hello` and `posts/path2/hello` things will break for them at the moment, unless they explicitly override links.

blog/content/docs/advanced.adoc
```diff
@@ -250,6 +250,11 @@ TIP: Those links are also available in the Qute data to allow <<roq-url>>.
 | The raw file path of the page without the extension.
 | `My$`, `my car` or `été/2024`
 
+| All
+| `:dir`
+| The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root.
```
ia3andy, line 255: shouldn't we have a small note for the draft dir?

ia3andy (approved):
LGTM

blog/content/docs/advanced.adoc
holly-cummins: I wondered about that too, but the behaviour is consistent with how we strip draft dirs for path. Let me look at the docs again to see where we document `draft` and how general it is.

holly-cummins: Actually, I had it wrong! The draft stripping was new in this PR, so I'll take it back out, and we can discuss in a separate PR or issue. There's an inconsistency between collections and not-collections. In the current Roq behaviour (before this PR), normal pages will always show `drafts` in the links of drafts, and drafts in collections will never have `drafts` in the URL. 

That could be confusing. I'd expect the preferred behaviour would be consistency between the two kinds of pages, and probably defaulting to 'show me my draft on the final url, same as if I'd used `draft: true`.'

With the new `:dir` placeholder, there's three possible behaviours:
- normal page: `drafts` is always in the url
- collection, with default link: `drafts` is never in the url
- collection, with `:dir`: `drafts` is always in the url

So I think stripping drafts is probably a good thing to do before this PR, and with this PR, there's an extra argument for it. 

I've raised https://github.com/quarkiverse/quarkus-roq/issues/1026.

holly-cummins:
This is now good to go, @ia3andy

ia3andy:
@holly-cummins conflict

holly-cummins:
> @holly-cummins conflict

Rebased, hopefully good now!

jtama:
I think we should all acknowledge now that @all-contributors please add @holly-cummins for code.

ia3andy (approved):
LGTM

PR #1020, opened 2026-06-29
Allow link defaults to be configured globally

holly-cummins:
At the moment link defaults are hardcoded in `LinkTemplate`, but it would be nicer if they were configurable. It would also make some of the decisions in #1013 easier, because users can make global changes if the default we choose is wrong for them. 

This is a 'prequel' to #1013, and I'm assuming it will merge first. So it just replicates the current defaults and doesn't change them. It does change quite a few files, which is mostly documentation. I did (with some umming and erring) change the internalLink method to not accept a fallback. Half the callers still use a fallback, but the other half don't, and the extra useless argument just seemed kind of awkward.

### DIFF
--- a/blog/content/docs/advanced.adoc
+++ b/blog/content/docs/advanced.adoc
@@ -312,11 +312,27 @@ NOTE: The slug derivation replaces all non-alphanumeric characters by `-` to mak
 
 Default link value:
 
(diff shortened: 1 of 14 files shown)


jtama (approved):
LGTM

PR #1021, opened 2026-06-29
Add list:whereExp filter for Jekyll where_exp migration

holly-cummins:
See discussion in https://github.com/quarkusio/quarkus/issues/54720

- Adds `list:whereExp` Qute template extension to handle Jekyll's `where_exp` filter, supporting comparison operators (`==`, `!=`, `>`, `>=`, `<`, `<=`, `contains`) with single or multiple conditions (AND logic)
- Preserves `JsonArray` return type so downstream methods like `.groupBy()` resolve correctly
- Makes `JekyllFiltersExtension` compilable, so that we can run tests (which this adds). To make it work, I also added Qute and Vert.x dependencies in the migration pom

### DIFF
--- a/migration/pom.xml
+++ b/migration/pom.xml
@@ -38,6 +38,18 @@
             <artifactId>jackson-dataformat-yaml</artifactId>
         </dependency>
 
+        <!-- Qute template extensions -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-qute</artifactId>
+        </dependency>
+
+        <!-- Vert.x JSON (JsonObject, JsonArray) -->
+        <dependency>
+            <groupId>io.vertx</groupId>
+            <artifactId>vertx-core</artifactId>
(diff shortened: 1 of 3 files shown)


ia3andy:
I think it's nice even out of jekyll, maybe we could split the RoqTemplateExtension to have one for list and put it there?

holly-cummins:
> I think it's nice even out of jekyll, maybe we could split the RoqTemplateExtension to have one for list and put it there?

I'm adding quite a few things into the `JekyllFilterExtension`, so what I was thinking would make sense is to do them all, and then once the dust has settled we can go through them all and pick out the ones we think should be made centrally available, and move them. That way nothing's blocked in the migration work waiting for releases. 

But +1 that making a grouped set of list extensions makes sense.

ia3andy:
Should I merge?

holly-cummins:
> Should I merge?

Yes please, thanks!

PR #1023, opened 2026-06-30
fix: handle null collections in getPosts method

### LINKED ISSUE #942: Editor NPE when collection is not configured
## Description

When opening the Roq Editor on a freshly created project (via `roq create`), the editor crashes with a NullPointerException when trying to list posts.

## Steps to reproduce

```bash
roq create the-missing-roq-verif-3
cd the-missing-roq-verif-3
roq start
# Press 'm' to open editor
```

## Error

```
java.lang.NullPointerException: Cannot invoke "io.quarkiverse.roq.frontmatter.runtime.model.RoqCollection.stream()" because the return value of "io.quarkiverse.roq.frontmatter.runtime.model.RoqCollections.get(String)" is null
        at io.quarkiverse.roq.editor.runtime.devui.RoqEditorJsonRPCService.getPosts(RoqEditorJsonRPCService.java:78)
```

## Environment

- Quarkus 3.35.4
- Roq (latest from `roq create`)
- JVM mode, dev profile

## Analysis

`RoqEditorJsonRPCService.getPosts()` calls `collections.get(collectionName)` without null-checking the result. If the collection is not configured or not yet resolved, it returns null and `.stream()` throws NPE.

matheusandre1:
Closes: #942

### DIFF
--- a/roq-editor/runtime/src/main/java/io/quarkiverse/roq/editor/runtime/devui/RoqEditorJsonRPCService.java
+++ b/roq-editor/runtime/src/main/java/io/quarkiverse/roq/editor/runtime/devui/RoqEditorJsonRPCService.java
@@ -75,7 +75,8 @@ public class RoqEditorJsonRPCService {
 
     @Blocking
     public List<PageSource> getPosts() {
-        return site.collections().get("posts").stream()
+        return Optional.ofNullable(site.collections().get("posts")).stream()
+                .flatMap(s -> s.stream())
                 .sorted(Comparator.comparing(Page::date, Comparator.nullsLast(Comparator.naturalOrder())).reversed())
                 .map(p -> new PageSource(p.collectionId(), p.sourcePath(), p.title(), p.description(), p.url().path(),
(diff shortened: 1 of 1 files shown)


ia3andy (approved):
Thanks!

roq-editor/runtime/src/main/java/io/quarkiverse/roq/editor/runtime/devui/RoqEditorJsonRPCService.java
```diff
@@ -75,7 +75,8 @@ public class RoqEditorJsonRPCService {
 
     @Blocking
     public List<PageSource> getPosts() {
-        return site.collections().get("posts").stream()
+        return Optional.ofNullable(site.collections().get("posts")).stream()
+                .flatMap(s -> s.stream())
```
jtama, line 79:
```suggestion
                .flatMap(RoqCollection::stream)
```

That would be clearer in my opinion.

PR #1024, opened 2026-06-30
fix: Enhance image handling for absolute paths in Page model

### LINKED ISSUE #935: Dir page frontmatter image: cannot reference public images
For directory-format pages (index.md with co-located files), the `image:` frontmatter field can only reference page-local images, not public/site-wide images.

**Root cause:** `Page.image(name)` checks `usePublicFiles()` which returns `false` for directory pages, so it only looks in page files. `isFullPath()` only matches `http://` URLs.

**Suggested fix:** In `Page.image(name)`, if the path starts with `/`, strip the leading slash and delegate to `site().image()`. For example, `image: /photo.jpg` would call `site().image("photo.jpg")`.

matheusandre1:
Fixes: #935


ia3andy (approved):
Thanks!

PR #1027, opened 2026-06-30
Add dir segment placeholders

holly-cummins:
Builds on #1013, so should be reviewed after that merges.

### DIFF
--- a/blog/content/docs/advanced.adoc
+++ b/blog/content/docs/advanced.adoc
@@ -6,6 +6,7 @@ include::../../includes/edit-contrib.adoc[]
 
 [#built-in-features]
 == Built-in features
+== Built-in features
 
 Roq and its themes include several built-in features for SEO, feeds, and discoverability. The link:/theme/base-theme/[base theme] provides the essential tags (SEO, favicon, Web Bundler), while the link:/theme/default-theme/[default theme] builds on top of it with a full layout, dark mode, sidebar, and more. If you are using either theme, most of these features are already enabled. Configure them through your site index frontmatter or by adding content files.
 
@@ -65,7 +66,7 @@ Just by doing so, Roq will generate a bunch of pages based on the pagination set
 NOTE: the first page uses the declaring page link.
 
 
-You can further customize your pagination by setting the page size and link format:
+You can further customize your pagination by setting direpage size and link format:
 
 [source,yaml]
 ----
@@ -252,7 +253,7 @@ TIP: Those links are also available in the Qute data to allow <<roq-url>>.
 
 | All
 | `:dir`
-| The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root. | The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root. | The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root. Use :dir[N] to skip the first N path segments — for example, :dir[2] drops the first two folders.
+| The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root. Use :dir[N] to skip the first N path segments — for example, :dir[2] drops the first two folders.
 | `posts`, `posts/v2/guides`, or empty
 
 
--- a/blog/content/docs/advanced.adoc
+++ b/blog/content/docs/advanced.adoc
@@ -6,7 +6,6 @@ include::../../includes/edit-contrib.adoc[]
 
 [#built-in-features]
 == Built-in features
-== Built-in features
 
 Roq and its themes include several built-in features for SEO, feeds, and discoverability. The link:/theme/base-theme/[base theme] provides the essential tags (SEO, favicon, Web Bundler), while the link:/theme/default-theme/[default theme] builds on top of it with a full layout, dark mode, sidebar, and more. If you are using either theme, most of these features are already enabled. Configure them through your site index frontmatter or by adding content files.
 
@@ -66,7 +65,7 @@ Just by doing so, Roq will generate a bunch of pages based on the pagination set
 NOTE: the first page uses the declaring page link.
 
 
-You can further customize your pagination by setting direpage size and link format:
+You can further customize your pagination by setting the page size and link format:
 
 [source,yaml]
 ----
@@ -253,7 +252,7 @@ TIP: Those links are also available in the Qute data to allow <<roq-url>>.
 
 | All
 | `:dir`
-| The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root. Use :dir[N] to skip the first N path segments — for example, :dir[2] drops the first two folders.
+| The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root. Use :dir[N] to skip the first N path segments — for example, :dir[2] drops the first two folders, giving a link of `my-post/my-file` for content in `/content/posts/2025/my-post/my-file.md`.
 | `posts`, `posts/v2/guides`, or empty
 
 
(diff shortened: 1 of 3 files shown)


ia3andy:
now conflict :)

holly-cummins:
> now conflict :)

Yeah, this will need rebasing to rebase onto the rebased #1013. :) Will do it now.

holly-cummins:
Ok, should be a small civilised diff now, and no conflicts.

roq-frontmatter/deployment/src/test/java/io/quarkiverse/roq/frontmatter/deployment/util/TemplateLinkTest.java
```diff
@@ -154,6 +154,7 @@ void testValidPlaceholdersNoException() {
         assertDoesNotThrow(() -> pageLink("", "/:path:ext", data));
         assertDoesNotThrow(() -> pageLink("", "/:name/:Slug", data));
         assertDoesNotThrow(() -> pageLink("", "/:dir/:slug", data));
+        assertDoesNotThrow(() -> pageLink("", "/:dir[1:]/:slug", data));
```
ia3andy: @holly-cummins this is the `[1:]` I was referring too

holly-cummins: Oh weird, I searched the github review page for `:1` and no results. I agree it looks odd and should be `:dir[1]`, will fix.

holly-cummins: On the other hand, it passes as it is now, which is also kind of surprising.

holly-cummins: Ohhh, they're everywhere, except the docs. But they are in the regexp. I think `dir[1]` is cleaner so I'll do that, do you agree?

holly-cummins: Fixed and squashed. Good spot, thanks! There are so many colons in the placeholders I think I went colon-blind and didn't spot those ones. :(

blog/content/docs/advanced.adoc
```diff
@@ -252,7 +252,7 @@ TIP: Those links are also available in the Qute data to allow <<roq-url>>.
 
 | All
 | `:dir`
-| The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root.
+| The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root. | The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root. | The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root. Use :dir[N] to skip the first N path segments — for example, :dir[2] drops the first two folders.
```
ia3andy: How does that look? it seems very long

ia3andy:
@holly-cummins I think doc is broken:
image

holly-cummins:
Yeah, that looks like a *very* bad merge, doesn't it? Will fix.

Edit: Now looking ok. Good spot, thanks!

the page&#39;s file path,

roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/utils/TemplateLink.java
```diff
@@ -188,6 +189,22 @@ public static String resolvePattern(String pattern, Map<String, String> values)
...
+                String[] segments = fullDir.isEmpty() ? new String[0] : fullDir.split("/");
+                String sliced = fromIndex >= segments.length ? ""
+                        : String.join("/", java.util.Arrays.copyOfRange(segments, fromIndex, segments.length));
+                result = result.replace(dirSliceMatcher.group(), sliced);
+                dirSliceMatcher = DIR_SLICE_PATTERN.matcher(result);
+            }
+        }
+
         for (Map.Entry<String, Supplier<String>> entry : mapping.entrySet()) {
             String key = entry.getKey();
```
ia3andy, line 209: A suggestion on the implementation in `TemplateLink.java`: since most templates won't use `:dir[N]`, a quick `result.contains(":dir[")` guard before compiling the regex matcher would skip the regex engine in the common case. And `dirSupplier.get()` could be called once before the while-loop and cached in a local variable, rather than re-invoked on each match.

holly-cummins: I never know how the regex engine handles these things under the hood. The regex engine should short circuit, right? And if it does, then doing a search with contains wouldn't be more efficient than doing a search with the regex engine (maybe?), and on the 'match' case (which I agree is rare) it would be two searches, and more costly. I just did a quick check, and https://stackoverflow.com/questions/28208793/in-java-which-is-faster-string-containssome-text-or-regex-that-looks-for says, basically, "not sure, but don't prematurely optimise," and https://sigpwned.com/2024/01/20/in-java-strings-much-faster-than-regexes-even-for-same-operations/ says string operations are faster than regex, as you say. But they're running loads of them in a tight loop. In this case, given this is unlikely to be a particularly hot path, I'd be inclined to prefer readability and maintainabilty – having the extra line of code and guard hurts clarity, and repeating the ":dir" placeholder string in two places is a bit non-DRY. The JIT people always say that idiomatic code compiles to the most performant code, in general, because those are the things the JIT is optimised for. 

Happy to move the .get() out of the loop, though, will do that now.

blog/content/docs/advanced.adoc
```diff
@@ -65,7 +66,7 @@ Just by doing so, Roq will generate a bunch of pages based on the pagination set
 NOTE: the first page uses the declaring page link.
 
 
-You can further customize your pagination by setting the page size and link format:
+You can further customize your pagination by setting direpage size and link format:
```
jtama: this is expected ? I mean `direpage` size ?

blog/content/docs/advanced.adoc
```diff
@@ -6,6 +6,7 @@ include::../../includes/edit-contrib.adoc[]
 
 [#built-in-features]
 == Built-in features
+== Built-in features
```
jtama: that's duplicate

roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/utils/TemplateLink.java
holly-cummins: Actually, I see the ":dir" constant is duplicated already. 

I'm running tests for the `.get()` out of the loop, but I'm not totally sure if it's idempotent. Hopefully the tests will confirm one way or another.

blog/content/docs/advanced.adoc
```diff
@@ -252,7 +253,7 @@ TIP: Those links are also available in the Qute data to allow <<roq-url>>.
 
 | All
 | `:dir`
-| The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root.
+| The directory portion of the page’s file path, slugified. For index files (directory pages), this is the grandparent directory (since the slug already captures the directory name). Empty for files at the content root. Use :dir[N] to skip the first N path segments — for example, :dir[2] drops the first two folders.
```
jtama: I think I'd like a sample of what it means. like for a content under /content/posts/2025/06/my-post/ => :dir[3] will result in blablabla

blog/content/docs/advanced.adoc
holly-cummins: Argh, my IDE was freezing when I did that edit. Will fix, sorry!

blog/content/docs/advanced.adoc
holly-cummins: I didn't want to make the text in the table too long, but I agree a sample is needed. It felt wrong to put one outside the table so I did nothing, but I'll put one in the table.

blog/content/docs/advanced.adoc
holly-cummins: The merge strikes again :(
Thanks!

ia3andy:
@holly-cummins ready to merge?

holly-cummins:
> @holly-cummins ready to merge?

Good to go!

PR #1028, opened 2026-06-30
Honour slug in frontmatter with sneaky path conditional

### LINKED ISSUE #1009: User overrides of slug is only honoured for collections, not normal pages
See reproducer in https://github.com/quarkiverse/quarkus-roq/pull/1008

The [docs](https://iamroq.dev/docs/advanced/#link-placeholders) do imply slug properties are supported, since `:slug` is defined as 
> The slugified title of the page, derived from the title. Defaults to the slug property in data, if available or using the slugified title, falling back to the name.

holly-cummins:
Resolves #1009. 

I've gone for the option of putting the condition in the `:path` resolution. Where there's an index file, the slug replaces the last folder, not the `index` file part (which isn't used in `:path`).

### DIFF
--- a/blog/content/docs/advanced.adoc
+++ b/blog/content/docs/advanced.adoc
@@ -242,7 +242,7 @@ TIP: Those links are also available in the Qute data to allow <<roq-url>>.
 
 | All
 | `:path`
-| The file path of the page, slugified (converted to a URL-friendly format) without the extension.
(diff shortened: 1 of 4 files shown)


ia3andy (commented):
Might just need to check attachements in dir (have a test for both page and documents when using :path)

ia3andy (commented):
sorry, Claude not listening and posting ..

ia3andy (approved):
Thanks!

PR #1034, opened 2026-07-02
feat: implement draft detection for substring paths

### LINKED ISSUE #789: Issue with draft dir detection
> `relativePath().contains(config.draftDirectory() + "/")` can misclassify paths where the draft directory name is only a substring of another segment (e.g. `posts/mydrafts/foo.html` contains `drafts/`). Consider matching on path segment boundaries (e.g., wrap both sides with `/` or split on `/` and compare segments) so only an actual `.../drafts/...` directory triggers the fallback. 

 _Originally posted by @Copilot in [#766](https://github.com/quarkiverse/quarkus-roq/pull/766/changes#r2882344792)_

matheusandre1:
Closes: #789


ia3andy:
@holly-cummins can you have a look, it's close to your recent work

roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep0SetupProcessor.java
```diff
@@ -31,6 +31,7 @@ public class RoqFrontMatterStep0SetupProcessor {
 
     // Data modifications are callbacks that transform front matter data during Step2 (assemble).
     // They run on every page/layout and can add/modify front matter keys.
+    private static final String DRAFT_DIRECTORY = "/";
```
holly-cummins: This variable naming is confusing to me. It's not the draft directory, because that comes from config. From the usage, it's intended to be the URL path separator, is that right? We should probably label the constant that way, or perhaps even re-use an existing constant. (Normally I'd use `File.separator` but here I think it's a url so always forward slashes. Not sure if there's an equivalent constant for that.

roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep0SetupProcessor.java
```diff
@@ -52,8 +53,10 @@ void amendDraftContent(RoqSiteConfig config,
         // Pages inside the draft directory are automatically marked as drafts
         // (unless they already have an explicit "draft" key in their front matter)
         dataModificationProducer.produce(new RoqFrontMatterDataModificationBuildItem(sourceData -> {
+            var relativePath = sourceData.relativePath();
             var isInDraftDirectory = sourceData.isPage() && sourceData.collection() != null
-                    && sourceData.relativePath().contains(config.draftDirectory() + "/");
+                    && (relativePath.startsWith(config.draftDirectory() + DRAFT_DIRECTORY)
```
holly-cummins: What happens if a user sets something like 

```
draft.directory=frogs/
```

with a trailing slash? Do we normalise elsewhere to avoid the trailing `//`? Might be worth adding some tests.

roq-frontmatter/deployment/src/test/java/io/quarkiverse/roq/frontmatter/deployment/apptest/RoqFrontMatterDraftSubstringTest.java
```diff
@@ -0,0 +1,29 @@
+package io.quarkiverse.roq.frontmatter.deployment.apptest;
```
holly-cummins: These two tests seem quite similar. I'd collapse them into a single test, with an `@Nested` for the two scenarios.

roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep0SetupProcessor.java
holly-cummins: Actually, reading the code, I think at this point it's representing a disk path, so maybe it _would_ be File.separator. Checking the rest of the code, we use `File.separator` one place and hardcoded "/" most places. Probably we mix the two and need a windows CI to make sure we haven't got wrong-slash-problems, but that's a separate issue!

holly-cummins (changes requested):
Functionally, this look good to me, and is a good thing to fix. I think there's some edge cases that might need testing, and also some renaming of constants and re-organising tests.

roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/RoqFrontMatterStep0SetupProcessor.java
holly-cummins: I raised https://github.com/quarkiverse/quarkus-roq/issues/1035 to track running CI on Windows.

holly-cummins (approved):
Looks perfect, thanks :)

PR #1039, opened 2026-07-02
Update Jekyll frontmatter converter for pagination and permalink migration

### LINKED ISSUE #985: Need to handle posts with no frontmatter (in migration or in roq)
Raising this so I don't lose it. If there are contents in `posts` without frontmatter, Jekyll silently ignores them. This makes it a convenient way to do includes. 

In Roq, the posts get published, and look silly, because they have no date. 

Either the converter (#979) or roq itself need to handle this. I haven't decided yet which is better, but I'm leaning to doing it in the converter.

holly-cummins:
Better handling of permalinks and redirects than before. 

Also resolves #985.


ia3andy (commented):
LGTM

PR #1041, opened 2026-07-03
feat: add isActive and nav methods to RoqUrl for improved URL handling

matheusandre1:
Closes: #805 

add new two methods to `RoqUrl` for  elegant active navigation

### DIFF
--- a/roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/model/RoqUrl.java
+++ b/roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/model/RoqUrl.java
@@ -218,6 +218,33 @@ public record RoqUrl(
     }
 
     /**
+     * Check if this url path matches the given path exactly (ignoring trailing slash).
+     *
+     * @param path the path to check against (including root path if configured)
+     */
+    public boolean isActive(String path) {
+        return normalize(path()).equals(normalize(path));
+    }
+
+    /**
+     * Returns "active" if this url matches one of the given paths, an empty string otherwise.
+     *
+     * @param paths the paths to check against (including root path if configured)
+     */
+    public String nav(String... paths) {
+        for (String path : paths) {
+            if (isActive(path)) {
+                return "active";
+            }
+        }
+        return "";
+    }
+
+    private static String normalize(String path) {
+        return path.endsWith("/") ? path.substring(0, path.length() - 1) : path;
+    }
+
+    /**
      * Remove the first occurrence of the given string from the path
      *
      * @param str the string to remove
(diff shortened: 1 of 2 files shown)


roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/model/RoqUrl.java
```diff
@@ -217,6 +217,45 @@ public RoqUrl replaceAll(String regex, String replacement) {
...
+     * Returns true if the current url path matches the given path exactly,
+     * or starts with the given prefix followed by {@code /}.
+     * This uses the same path semantics as {@link #startsWith(String)}.
+     * <p>
+     * Example: "{#if page.url.isActive('/docs')}" → true for /docs, /docs/, /docs/getting-started
+     *
+     * @param path the path to check against (including root path if configured)
+     * @return true if this url is active for the given path
+     */
+    public boolean isActive(String path) {
```
jtama, line 225: Ok so we want exact match only, really simple feature. Maybe with path normalization only.

Also there twice as much comments than code. That is really noisy.

jtama (changes requested):

jtama (approved):

jtama:
@matheusandre1  Did you find any use of this in the blog ? 
If yes, could you provide a second PR ?

roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/model/RoqUrl.java
```diff
@@ -217,6 +217,33 @@ public RoqUrl replaceAll(String regex, String replacement) {
...
+        return normalize(path()).equals(normalize(path));
+    }
+
+    /**
+     * Returns "active" if this url matches one of the given paths, an empty string otherwise.
+     *
+     * @param paths the paths to check against (including root path if configured)
+     */
+    public String nav(String... paths) {
```
ia3andy, line 234: @matheusandre1 @jtama  why not use isActive(String path, String... paths)?

I am not sure why we name it nav?

roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/model/RoqUrl.java
```diff
@@ -217,6 +217,33 @@ public RoqUrl replaceAll(String regex, String replacement) {
...
+        for (String path : paths) {
+            if (isActive(path)) {
+                return "active";
+            }
+        }
+        return "";
+    }
+
+    private static String normalize(String path) {
+        return path.endsWith("/") ? path.substring(0, path.length() - 1) : path;
```
ia3andy, line 244: Can use StringPath utils instead

ia3andy (commented):
Could you fix those @matheusandre1 (in a follow up PR)

PR #1052, opened 2026-07-04
ci: windows preview

### LINKED ISSUE #1035: Should run tests on Windows on CI
The codebase uses "/" in most places, and File.separator in one or two places. On Linux and Mac, those two are the same, but for Windows they are not. To make matters worse, in a filesystem-to-web context, sometimes you want the File.separator, and sometimes you want "/". The `GitContentFilter` does some normalisation, but I can't see it elsewhere. 

It would be worth having a regular CI run on Windows to make sure issues don't creep in. Obviously, that would be as well as Linux, not instead of Linux. :)

matheusandre1:
Closes: #1035

The only explicit normalization in our codebase so far is this: [here](https://github.com/quarkiverse/quarkus-roq/blob/157abfa80be81ac3ffcb526a9b6c82dffe152135/roq-editor/deployment/src/main/java/io/quarkiverse/roq/editor/deployment/git/GitContentFilter.java#L77)

I haven't seen any more `File.separator` instances so far.

### DIFF
--- a/.github/workflows/build-pr-preview.yml
+++ b/.github/workflows/build-pr-preview.yml
@@ -9,9 +9,41 @@ on:
       - '*.md'
       - '*.txt'
       - '.all-contributorsrc'
+permissions:
+  contents: read
+  pull-requests: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
+  test:
+    name: Test on ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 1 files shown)


ia3andy (changes requested):
Test is another workflow

ia3andy:
Should be on build workflow

matheusandre1:
> Should be on build workflow

but, [here](https://github.com/quarkiverse/quarkus-roq/blob/5e8ffcffc168fc552cf44b6b6825e99ca869df24/.github/workflows/build.yml#L44) It's already at the headquarters.

ia3andy:
We already have windows CI for PR it is not needs

PR #1062, opened 2026-07-06
feat: Add RSS feed support in roq-base theme layout

### LINKED ISSUE #1058: Add {#rss} tag to the base theme default layout head
The base theme's `default.html` layout does not include the `{#rss site /}` tag in its `<head>`. Users who create an RSS feed with `content/rss.xml` still need to manually override the layout to add the RSS auto-discovery link.

Since the tag is a no-op when no RSS file exists, it could be included by default in the base theme's head section alongside `{#seo /}` and `{#favicon /}`.

matheusandre1:
Closes: #1058

### DIFF
--- a/roq-frontmatter/runtime/src/main/resources/templates/theme-layouts/roq-base/default.html
+++ b/roq-frontmatter/runtime/src/main/resources/templates/theme-layouts/roq-base/default.html
@@ -6,6 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   {#seo page site/}
   {#favicon site/}
+  {#rss site/}
   {#bundle /}
   {/insert}
   {#insert head /}


roq-frontmatter/deployment/src/test/resources/base-theme-site/content/rss.xml
```diff
@@ -0,0 +1 @@
+{#include fm/rss.html /}
```
ia3andy: remove this file, this is opt-in

ia3andy (changes requested):
The rss tag will be added in head if it's enabled. We don't want to add it if the user don't want it.

ia3andy:
@matheusandre1 you are just using IA blindly here, please give it a bit of thought, go back to your previous code. Just remove the rss file from the commit. Opt-in is not via config but via the file, if you add it you don't give the user a choice

matheusandre1:
> @matheusandre1 you are just using IA blindly here, please give it a bit of thought, go back to your previous code. Just remove the rss file from the commit. Opt-in is not via config but via the file, if you add it you don't give the user a choice

i wish it were AI, it just an autistic person thinkking out loud always and writing code

ia3andy (approved):
LGTM

PR #1078, opened 2026-07-10
Fix unbalanced quote in aliases plugin redirect template

holly-cummins:
The redirect template had quotes around  the`{url}` placeholder which caused double-quoting when the URL was substituted, breaking the meta refresh tag. Actually, it's worse than that, because it didn't have double-quotes, it had *a* double quote (singular). :)

Changed: `content="0; url="{url}`
To: `content="0; url={url}"`

```
  // Original (buggy):
  <meta http-equiv="refresh" content="0; url="{url}">
     
  // After substitution with https://...:
  <meta http-equiv="refresh" content="0; url="https://...">
```

  The second " prematurely closes the content attribute, leaving https://...">
  as naked text outside any attribute. 

  With the fix:
```
  // Fixed:
  <meta http-equiv="refresh" content="0; url='{url}'">
  
  // After substitution:
  <meta http-equiv="refresh" content="0; url='https://...'">
```


jtama (approved):
LGTM

ia3andy (approved):
Good catch, weird that it was still working 😅

PR #1079, opened 2026-07-10
Add site.tags support

### LINKED ISSUE #964: Add site.tags support to roq-tagging
I spotted this gap when converting quarkus.io to Roq.

  Goal: Make site.tags work like Jekyll by aggregating tags across all collections.

  Implementation:

  1. Create extension method in roq-plugin-tagging:

 ```
 // roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java

  @TemplateExtension
  public class RoqTaggingTemplateExtension {

      // Existing method
      public static List<TagCount> tagsCount(RoqCollection collection) {
          // ... existing implementation
      }

      // NEW: Add site.tags support (Jekyll compatibility)
      public static Map<String, List<Page>> tags(Site site) {
          // Aggregate tags across ALL collections
          Map<String, List<Page>> allTags = new LinkedHashMap<>();

          for (RoqCollection collection : site.collections().values()) {
              // For each collection, get its documents
              for (NormalPage page : collection.pages()) {
                  // Extract tags from page frontmatter
                  Object tagsObj = page.data().get("tags");
                  List<String> tags = extractTags(tagsObj);

                  for (String tag : tags) {
                      allTags.computeIfAbsent(tag, k -> new ArrayList<>()).add(page);
                  }
              }
          }
  
          return allTags;
      }

      private static List<String> extractTags(Object tagsObj) {
          if (tagsObj == null) return Collections.emptyList();
          if (tagsObj instanceof List) {
              return ((List<?>) tagsObj).stream()
                  .map(Object::toString)
                  .collect(Collectors.toList());
          }
          if (tagsObj instanceof String) {
              // Handle comma-separated tags
              return Arrays.asList(((String) tagsObj).split("\\s*,\\s*"));
          }
          return Collections.emptyList();
      }
  }
```

  2. Usage in templates:

```
  {! Jekyll-compatible tag listing !}
  {#let tag_words=site.tags.sort}
  {#for entry in tag_words}
    {#let tag=entry.key}
    {#let posts=entry.value}
    <a href="/blog/tag/{=tag.toLowerCase}">{=tag}</a> ({=posts.size})
    {/let}{/let}
  {/for}
  {/let}
```

When this is done, `makeSiteTagsLenient()` can be removed from the Liquid converter.

  Pros:
  - Full Jekyll compatibility
  - No manual migration needed
  - Works across all Roq sites using tagging plugin

  Cons:
  - Aggregates ALL collections (could be expensive on large sites)
  - Adds API surface to maintain
  - May not match how Roq's collection-first model should work

### PR DESCRIPTION
**Pull request by holly-cummins**
Resolves #964

### DIFF
--- a/roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java
+++ b/roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java
@@ -3,10 +3,20 @@ package io.quarkiverse.roq.plugin.tagging;
 import static java.util.stream.Collectors.counting;
 import static java.util.stream.Collectors.groupingBy;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
 
+import io.quarkiverse.roq.frontmatter.runtime.RoqTemplateExtension;
+import io.quarkiverse.roq.frontmatter.runtime.model.Page;
 import io.quarkiverse.roq.frontmatter.runtime.model.RoqCollection;
+import io.quarkiverse.roq.frontmatter.runtime.model.Site;
 import io.quarkus.qute.TemplateExtension;
+import io.vertx.core.json.JsonArray;
 
 @TemplateExtension
 public class RoqTaggingTemplateExtension {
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by holly-cummins** [comment]
I've updated this to be both shorter, and handle spaces correctly. Good spot, thanks @ia3andy!

Sadly, in extending the testing, I tripped over #1086, so I've had to disable one of the tests.

File: roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java [hunk 1]
```diff
@@ -42,4 +47,41 @@ public static List<TagCount> tagsCount(RoqCollection collection) {
...
+     * Tag: {entry.key} has {entry.value.size} pages
+     * {/for}
+     *
+     * Or with sorting:
+     * {#let tag_words=site.tags.entrySet.sort('key')}
+     * {#for entry in tag_words}
+     * <a href="/blog/tag/{entry.key}">{entry.key}</a> ({entry.value.size})
+     * {/for}
+     * {/let}
+     *
```
jtama, line 66: I feel this should be in the documentation rather than the Javadoc.

roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java
```diff
@@ -42,4 +47,41 @@ public static List<TagCount> tagsCount(RoqCollection collection) {
...
+     * @param site the Roq site
+     * @return a map of tag names to lists of pages with that tag
+     */
+    public static Map<String, List<Page>> tags(Site site) {
+        if (site == null) {
+            return Map.of();
+        }
+
+        Map<String, List<Page>> tagMap = new LinkedHashMap<>();
```
jtama, line 76: Out of curiosity: Why a LinkedHashMap ? Is there any reason to keep track of the tag order ?

roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java
holly-cummins: That's a good point, will move it.

roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java
holly-cummins: Good question. I guess they'd usually be sorted when they're rendered, but that would usually be explicitly done by the consuming code. I know (from digging into #1086) that `allPages` does not have a deterministic order, and I guess if that's the case, there's no reason for the consumers of it to be fussy about order. Will update.

holly-cummins: Although actually, this got pulled into `main` with #1086, so after rebasing, this PR will be empty! I'll just use it for these two updates.

holly-cummins: On discussion with @ia3andy, his preference is that things should be ordered where possible, and order preserved. What came out of site allTags wasn't ordered, which made the LinkedHashMap a bit useless, so I've gone the other way and kept the linked map, but made sure allTags is ordered. This PR is a bit of a mess so I'll close it, and those changes can be a clearer diff in https://github.com/quarkiverse/quarkus-roq/pull/1088.

holly-cummins:
Closing as merged in #1086.

PR #1082, opened 2026-07-13
Expose resolved layout in page.data when it comes from defaults

### LINKED ISSUE #1080: Expose collection-level layout in page.data.layout
When a collection defines a default layout via application.properties (e.g. `site.collections.guides.layout=guides`), Roq correctly uses that layout for rendering, but the value is not available via page.data.layout in templates. `page.data.layout` is only populated when a page explicitly declares `layout:` in its own frontmatter.
    
This matters for templates that need to vary behavior based on which layout a page uses. For example,  the `<body>` CSS class could be wrong. In quarkus.io guide pages render `<body 
  class="base">` instead of `<body class="guides">`, which breaks `body.guides 
  div.guide { ... }` CSS rules (TOC sidebar, text colors, padding, mobile
  word-wrap). 

This may affects other default-ed properties, so the fix should be generic.

holly-cummins:
Resolves https://github.com/quarkiverse/quarkus-roq/issues/1080

When a collection defines a default layout via config (e.g. site.collections.guides.layout=guide), the layout was used for rendering but page.data.layout returned null because only frontmatter-declared values were in the data JsonObject. Now processTemplate writes the resolved layout key back into data when neither layout nor theme-layout is present in frontmatter. 

I wondered if other defaults would need to be handled in the same way (or if the solution should be generic), but Claude says not: 

 Looking at the other ConfiguredCollection properties:
  - enabled — controls whether the collection is active, not a page property
  - hidden — controls whether pages are served on a path, not page-level data
  - future — controls visibility of future-dated posts, not page-level data
  - fromData — collection config plumbing, not page data

  The only other candidate is link, but it's a URL pattern template rather than a semantic
  property templates would query. Templates use page.url for the resolved URL. I don't think
  page.data.link returning the pattern template (/:collection/:slug/) would be useful —
  it's more of an internal routing concern.

  So layout is the only property that matters here. The issue mentions "other default-ed
  properties" as a possibility, but looking at the actual config, layout is the only one
  that's both a user-facing semantic property and has a collection/page-level default that
  templates would want to inspect.

### DIFF
--- a/roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/util/RoqFrontMatterAssembleUtils.java
+++ b/roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/util/RoqFrontMatterAssembleUtils.java
@@ -60,6 +60,10 @@ public final class RoqFrontMatterAssembleUtils {
         Optional<String> sourceTheme = extractSourceTheme(isThemeLayout, metadata.templateId());
         String layoutId = availableLayouts.resolveLayoutId(config.theme(), sourceTheme, layoutRef);
 
+        if (layoutId != null && !data.containsKey(LAYOUT) && !data.containsKey(THEME_LAYOUT)) {
+            data.put(LAYOUT, RoqFrontMatterLayoutUtils.getLayoutKey(config.theme(), layoutId));
+        }
+
         for (RoqFrontMatterDataModificationBuildItem modification : dataModifications) {
             data = modification.modifier()
                     .modify(new RoqFrontMatterDataModificationBuildItem.SourceData(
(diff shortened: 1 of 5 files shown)


jtama (approved):
Looks good to me, though we did add in the contributing guidelines that there must to be any `co-authored by <add your llm here>`.

Once it's commited, it's all yours.

holly-cummins:
> Looks good to me, though we did add in the contributing guidelines that there must to be any `co-authored by <add your llm here>`.
> 
> Once it's commited, it's all yours.

Oh rats, I'd noticed that and had been carefully reviewing and updating commits to make sure there were no co-authored. Will fix!

holly-cummins:
Attribution on commit fixed.

PR #1092, opened 2026-07-16
fix: improve Lunr search ranking and add URL to results

ia3andy:
## Summary

- **Fix search-boost scaling**: marketplace pages used `search-boost: 20`, which completely overrode keyword relevance. Reduced to `1.2` to stay within the BM25 scoring range where keyword matches can still win.
- **Fix section heading boost**: was inverted (h6 ranked above h2) and used large integer values. Now h2 gets +0.6, h6 gets +0.2.
- **Rebalance field-level boosts**: content was 100, title was 10. Rebalanced all fields to [1, 2] range so BM25 handles relevance naturally.
- **Add URL path to search results**: each result now shows the page path below the title.
- **Support float search-boost values**: frontmatter `search-boost` now accepts decimals (e.g., `1.2`).
- **Replace SCSS with plain CSS**: no SCSS features were used.
- **Document boost behavior**: explain how boost interacts with BM25 and recommend values between 0 and 2.

### DIFF
--- a/blog/content/marketplace/plugins/aliases.md
+++ b/blog/content/marketplace/plugins/aliases.md
@@ -6,7 +6,7 @@ icon: fa-solid fa-shuffle
 install-name: aliases
 tags: [navigation, seo]
 source: https://github.com/quarkiverse/quarkus-roq/tree/main/roq-plugin/aliases
-search-boost: 20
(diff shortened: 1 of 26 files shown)


jtama (approved):
LGTM

Issue #1086, opened 2026-07-15
Adding a tagging template causes duplicate pages in `Site.allPages`


holly-cummins:
If I add a tags template for tag index pages, duplicate copies of my site's pages get added into `Site.allPages`

<img width="183" height="52" alt="Image" src="https://github.com/user-attachments/assets/16560e8a-7f8a-4765-8271-eb4026fc58e4" />

The problem seems to happen in this method:

```
    private static ArrayList<Page> getAllPages(List<NormalPage> pages, RoqCollections collections) {
        final ArrayList<Page> allPages = new ArrayList<>(pages);

        for (RoqCollection value : collections.collections().values()) {
            // Uh oh! There's a collection for each tag, but that collection includes copies of all the pages with that tag 
            allPages.addAll(value);
        }
        return allPages;
    }
```

Example debug output (hand-edited for clarity, so hopefully I didn't mess something up): 

```
adding posts/tag/cloud-native from ConfiguredCollection[id=posts/tag/cloud-native, derived=true, hidden=false, future=false, layout=post, link=/:collection/:slug/, fromData=Optional.empty]
collection contents posts/post-quarkus.md tags {"layout":"default","title":"Quarkus Post","tags":"quarkus, java, cloud-native","date":"2024-01-02"}

adding posts/tag/native from ConfiguredCollection[id=posts/tag/native, derived=true, hidden=false, future=false, layout=post, link=/:collection/:slug/, fromData=Optional.empty]
collection contents posts/post-spaces.md tags {"layout":"default","title":"Quarkus Post With Tags With Spaces","tags":"bunnies, some topic, cloud native","date":"2024-01-02"}

adding posts/tag/jvm from ConfiguredCollection[id=posts/tag/jvm, derived=true, hidden=false, future=false, layout=post, link=/:collection/:slug/, fromData=Optional.empty]
collection contents posts/post-java.md tags {"layout":"default","title":"Java Post","tags":["java","jvm"],"date":"2024-01-01"}

adding posts/tag/java from ConfiguredCollection[id=posts/tag/java, derived=true, hidden=false, future=false, layout=post, link=/:collection/:slug/, fromData=Optional.empty]
[...]

PR #1094, opened 2026-07-16
Convert tags to lowercase (when configured to do so) in tags aggregation

holly-cummins:
This fixes an issue spotted by @jtama. (Jekyll has a similar issue, but for different root causes.) It's a problem in #1079/#1086, in which the lowercase property is not honoured.

The `quarkus.roq.tagging.lowercase property` doesn't  work as expected:

  1. Slugification is case-preserving by default - RoqTemplateExtension.slugify() preserves case unless explicitly told to  lowercase
  2. The lowercase config only affects build-time tag page generation - It's applied in `RoqPluginTaggingProcessor` (lines 91-93) when creating tag pages
  3. Runtime site.tags aggregation ignores the config - The `RoqTaggingTemplateExtension.tags()` method doesn't apply the lowercase config,  so it returns both 'AI' and 'ai' as separate tags
  
  The fix ended up being a bit tricky, because the lowercase config is build time config, but it needs to be available at runtime for access by the extension. I've also added loads of tests, which reproduced the problem and now pass.

### DIFF
--- a/roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java
+++ b/roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java
@@ -8,6 +8,8 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.eclipse.microprofile.config.ConfigProvider;
+
 import io.quarkiverse.roq.frontmatter.runtime.model.Page;
 import io.quarkiverse.roq.frontmatter.runtime.model.RoqCollection;
 import io.quarkiverse.roq.frontmatter.runtime.model.Site;
@@ -20,8 +22,9 @@ public class RoqTaggingTemplateExtension {
      * Returns a list of all tags from the given collection, with each tag slugified.
      */
     public static List<String> allTags(RoqCollection collection) {
+        boolean lowercase = isLowercase();
         return collection.stream()
-                .flatMap(documentPage -> RoqTaggingUtils.slugifiedTagStringsStream(documentPage.data()))
+                .flatMap(documentPage -> RoqTaggingUtils.slugifiedTagStringsStream(documentPage.data(), lowercase))
                 .toList();
     }
 
@@ -30,9 +33,10 @@ public class RoqTaggingTemplateExtension {
      * each tag appears. Each tag is slugified.
      */
     public static List<TagCount> tagsCount(RoqCollection collection) {
+        boolean lowercase = isLowercase();
         return collection
                 .stream()
-                .flatMap(documentPage -> RoqTaggingUtils.slugifiedTagStringsStream(documentPage.data()))
+                .flatMap(documentPage -> RoqTaggingUtils.slugifiedTagStringsStream(documentPage.data(), lowercase))
                 .collect(groupingBy(tag -> tag, counting())).entrySet()
                 .stream().map(entry -> new TagCount(entry.getKey(), entry.getValue()))
                 .toList();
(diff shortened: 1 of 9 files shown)


roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java
```diff
@@ -60,15 +64,22 @@ public static Map<String, List<Page>> tags(Site site) {
...
-            for (String tag : RoqTaggingUtils.slugifiedTagStrings(page.data())) {
+            for (String tag : RoqTaggingUtils.slugifiedTagStrings(page.data(), lowercase)) {
                 tagMap.computeIfAbsent(tag, k -> new ArrayList<>()).add(page);
             }
         }
 
         return tagMap;
     }
 
+    private static boolean isLowercase() {
```
jtama, line 79: Seems to me this method could be to the RoqTaggingUtils. 

I mean, the best way to remove that boolean arguments from the `slugifiedTagStrings` method

roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java
```diff
@@ -60,15 +64,22 @@ public static Map<String, List<Page>> tags(Site site) {
...
+        return ConfigProvider.getConfig()
```
ia3andy, line 80: We are re-reading the value many times. I wonder if there is a cleaner way to achieve this for the static methods (if we could avoid hardcoding those).

@radcortez @mkouba any thoughts?

mkouba: Just cache the value in a static volatile field? It's a build-time property so it should not change at runtime.

ia3andy: It should also be possible to access it via the COnfig class using CDI right?

ia3andy: LazyValue is good for this

ia3andy: I'll merge and solve it in a follow up PR

mkouba: > It should also be possible to access it via the COnfig class using CDI right?

It is but I see no benefit there.

> LazyValue is good for this

Yep, `io.quarkus.qute.LazyValue` is a good fit.

holly-cummins:
I should add that this also affected code before #1086.

roq-plugin/tagging/runtime/src/main/java/io/quarkiverse/roq/plugin/tagging/RoqTaggingTemplateExtension.java
ia3andy: @mkouba not hardcoding the path to config and default value

mkouba: > @mkouba not hardcoding the path to config and default value

You can `@Inject RoqTaggingConfig` or perform lookup like `Arc.instance().instance(RoqTaggingConfig.class).get()`.

holly-cummins: My recollection is that my Claude tried both of those, and neither worked (to both our surprise). I think the @Inject didn't work because the method needs to be static. The Arc instance just ... didn't work, and Claude didn't produce output about why.

mkouba: > I think the @Inject didn't work because the method needs to be static.

Yes, that's expected. I mentioned `@Inject RoqTaggingConfig` just for completeness, it's a recommended approach for regular beans and instance methods.

> The Arc instance just ... didn't work, and Claude didn't produce output about why.

That's really weird, it should work for `BUILD_AND_RUN_TIME_FIXED`.

ia3andy: yeah need unremovable..

mkouba: Ah, yes, that's it. Programmatic lookup does not count when detecting unused beans... you'd need an `UnremovableBeanBuildItem` here, because neither `quarkus.arc.unremovable-types` nor `@Unremovable` can be used.

PR #1100, opened 2026-07-17
Handle newUrl and excerpt frontmatter in Jekyll conversion

holly-cummins:
I believe Jekyll to Roq conversion is now working well, so I'm trying to get code off my laptop into the repo. There's a lot of it, sorry! 

I'm trying to break it up into digestible chunks. 

This one is fairly painless – it just adds support for `newUrl` and `excerpt` properties in front matter, and maps them to `aliases:` and `description` properties.


jtama (commented):
The code is fine and the excerpt use case seems to be well covered. No tests for aliases though.

holly-cummins:
> The code is fine and the excerpt use case seems to be well covered. No tests for aliases though.

Oh you're quite right. Will add, thanks!

*Edit:* Added.

jtama (approved):

PR #1101, opened 2026-07-17
More support for Jekyll macros and also enable broader-scoped variables by mutation

holly-cummins:
More utilities to match capabilities that I found in Liquid templates. 

Most of these are just ... functions:

- sort
- group
- date handling 
- escaping html
- splitting strings 
- capitalizing strings

As with the other changes, I suggest we put them into the Jekyll template first, and if they look useful after the dust has settled, we move them to the central template. Otherwise conversion gets blocked on Roq release cadences.

The one function here which is semantically interesting is the mutable map. I used this to fix a nasty problem I kept hitting during the conversion. Qute variables are scoped to the block they're in, but Liquid variables have a higher scope (file, I think). Normally the converter could get round this by hoisting variables, but sometimes it just couldn't, usually when the value of the variable would get changed later on. So what I've done is, in rare cases, shoved the value into a map, and then used map reading and writing to make the variable higher-scoped and mutable. It's icky, but the patterns are there in Jekyll, and we want to get people able to convert.

### DIFF
--- a/migration/src/main/java/io/quarkus/tools/migration/JekyllFiltersExtension.java
+++ b/migration/src/main/java/io/quarkus/tools/migration/JekyllFiltersExtension.java
@@ -1,9 +1,19 @@
 package io.quarkus.tools.migration;
 
 import java.lang.reflect.Method;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.regex.Pattern;
 
 import io.quarkus.qute.RawString;
 import io.quarkus.qute.TemplateExtension;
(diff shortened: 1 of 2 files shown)


migration/src/main/java/io/quarkus/tools/migration/JekyllFiltersExtension.java
```diff
@@ -13,68 +23,87 @@
 @TemplateExtension
 public class JekyllFiltersExtension {
 
-    /**
-     * Null-safe get for JsonObject. Prevents NPE when key is null (e.g. from ?? operator).
-     */
+    // Null-safe get — prevents NPE when key is null (e.g. from ?? operator)
     static Object get(JsonObject obj, String key) {
-        if (obj == null || key == null || key.isEmpty()) {
+        if (obj == null || key == null || key.isEmpty())
```
jtama: All those ifs without braces kind of bother me.

I know we are not suppose to look at those details in the AI era, but still.

jtama (commented):
Honestly I just don't understand 80% of the code, so if it works on the quarkus.io migration process...

holly-cummins:
> Honestly I just don't understand 80% of the code, so if it works on the quarkus.io migration process...

I know - there's SO much code. :( 

I've been having to be really empirical in how I evaluate it ("does quarkus.io work a bit better now? ok, it's good!")

Once it's all in, I might attempt some refactorings to slim down the converter code and make it more readable, but when I do that, I'd want to keep the tests the same – so that means the output wouldn't get slimmed down, even if there are opportunities to do so. 

I've got the big integration test that tests end to end by doing an actual conversion (#1006, draft), but even once that's merged, it would only test e2e on the simple path, not the really complex cases that quarkus.io is hitting. I guess maybe I should try and beef it up with a few more complex things like broader-scoped or mutable variables.

migration/src/main/java/io/quarkus/tools/migration/JekyllFiltersExtension.java
holly-cummins: Agree, will improve the style of that one.

holly-cummins: Fixed! You were totally right, those braceless-ifs were risky.

jtama (approved):

PR #1103, opened 2026-07-17
Add JekyllLayoutConverter for CSS bundle tag migration

### LINKED ISSUE #1070: Conversion Jekyll to Roq: Fix needed to improve Lighthouse performance score
As with #1069, I'm not sure if this is a converter issue or a Roq issue, but since they live in the same repo, raising this here while I figure which code to blame. :)

The lighthouse reports performance ranking for quarkus.io dropped from [49](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1782923375367-10880.report.html) to [42](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1783347017042-92460.report.html) after conversion. There's so many issues in the original site it's hard to figure out the diff. I *think* all the audits are the same but it's just that the timings are worse, especially "total blocking time". That will make it hard to fix :( 

I also don't know how much volatility there is in the results, we will need more runs to gather a sense.

Some things got *better* post-conversion (the audits pass for Roq but not Jekyll):

Image

holly-cummins:
Fixes #1070 

The bulk of the slower performance was caused by two css files that couldn't be found, causing a 404. Some delay was also caused by having the `{#bundle}` right after `<head>`. This fixes all three of those issues, by 

- Putting the  {#bundle} declaration where the first original css import is
- Removing all the old css imports, since the bundle handles everything, and 404s are slow 

This eliminates 404 errors from broken static/bundle/app.css references that caused ~100-200ms Lighthouse TBT regression in Jekyll migrations.

Obviously, these classes need to be called, but that's another PR.

### DIFF
--- a/migration/src/main/java/io/quarkus/tools/JekyllLayoutConverter.java
+++ b/migration/src/main/java/io/quarkus/tools/JekyllLayoutConverter.java
@@ -1,0 +1,58 @@
+package io.quarkus.tools;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.StringReader;
+import java.io.StringWriter;
+import java.util.regex.Pattern;
+
+/**
+ * Converts Jekyll layout HTML to Roq layout HTML.
+ * Primarily handles replacing /assets/css/ stylesheet links with {#bundle /} tag.
+ */
+public class JekyllLayoutConverter {
+
+    private static final Pattern ASSETS_CSS_PATTERN = Pattern.compile(".*href=.*['\"].*?/assets/css/.*");
+
+    /**
+     * Replace the first /assets/css/ stylesheet link with {#bundle /} and remove
+     * any additional /assets/css/ links.
+     *
+     * This positions the web-bundler tag where Jekyll's main CSS was located,
+     * preserving the CSS load order while removing redundant references.
+     *
+     * Preserves:
+     * - Non-assets CSS (e.g., /guides/stylesheet/config.css)
+     * - External CDN CSS (e.g., https://use.fontawesome.com/...)
+     * - All other HTML
+     *
+     * @param content HTML content to convert
+     * @return Converted HTML with bundle tag replacing /assets/css/ links
+     * @throws IOException if reading the content fails
+     */
+    public String replaceAssetsCssWithBundle(String content) throws IOException {
+        StringWriter output = new StringWriter();
+        boolean bundleTagInserted = false;
+
+        try (BufferedReader reader = new BufferedReader(new StringReader(content))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                if (ASSETS_CSS_PATTERN.matcher(line).matches()) {
+                    if (!bundleTagInserted) {
+                        // Replace first /assets/css/ link with {#bundle /}
+                        // Preserve indentation from the original line
+                        String indentation = line.substring(0, line.indexOf('<'));
+                        output.write(indentation + "{#bundle /}\n");
+                        bundleTagInserted = true;
+                    }
+                    // Skip this line (first replacement already done, or additional link to remove)
+                } else {
+                    output.write(line);
+                    output.write('\n');
+                }
+            }
+        }
+
+        return output.toString();
+    }
+}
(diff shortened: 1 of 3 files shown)


migration/src/main/java/io/quarkus/tools/JekyllLayoutConverter.java
```diff
@@ -0,0 +1,58 @@
...
+            while ((line = reader.readLine()) != null) {
+                if (ASSETS_CSS_PATTERN.matcher(line).matches()) {
+                    if (!bundleTagInserted) {
+                        // Replace first /assets/css/ link with {#bundle /}
+                        // Preserve indentation from the original line
+                        String indentation = line.substring(0, line.indexOf('<'));
+                        output.write(indentation + "{#bundle /}\n");
+                        bundleTagInserted = true;
+                    }
+                    // Skip this line (first replacement already done, or additional link to remove)
```
jtama, line 48: It feels like an elegant way to remove duplication ❤️

migration/src/test/java/io/quarkus/tools/JekyllLayoutConverterTest.java
```diff
@@ -0,0 +1,173 @@
...
+        String result = converter.replaceAssetsCssWithBundle(input);
+
+        // Expected output
+        String expected = """
+                <head>
+                  <title>{{ page.title }}</title>
+                  <meta charset="utf-8">
+                  <link rel="canonical" href="{{ canonical_url | prepend: site.url }}">
+                  <link rel="shortcut icon" type="image/png" href="/favicon.ico" >
+                  <link rel="stylesheet" href="/guides/stylesheet/config.css" />
```
jtama, line 164: Just a question. Shouldn't that be also bundled ?

holly-cummins: Yeeees, it does feel like it should. Let me have a look.

holly-cummins: Oh, I know why! I think it's because the imaginary user in this test case chose to put their css in a weird place. And I think Roq only considers css in assets eligible for bundling.

holly-cummins: Er, I just double-checked, and it's not assets, it's:

  1. web/ directory (project root - what Roq usually uses)
  2. src/main/resources/web/app/ (standard Quarkus Web Bundler location, but Roq usually uses project root web/ instead)
  
  But in any case, not /content. 

Now, you could say the converter should be aggressive and move strangely-placed css to somewhere where it can get bundled, but I don't know if that would be what users would want.

jtama: mkay, so the use case is, if for some reason, user doesn't want it's css to be bundled. In which case we respect it's choice and make it work anyway.

shouldn't we emit a warning though ? Inviting the user to follow prescribed folder standards ?

holly-cummins: You're right, we should alert the user their choice is causing them to miss out. The only downside is the conversion runs are SO VERBOSE that any warning will be totally lost. But that's not a reason not to do it; we can warn now, and deal with verbosity later.

jtama: Ok so _maybe_, we leave it has is, but create an issue, as follow up for when we have a full jekyll to conversion.

We could have a file summary that gathers warning.

holly-cummins: That would be really nice. I'll raise an issue to track adding the right warnings, removing the wrong chattiness, and making a nice summary.

holly-cummins: I've raised https://github.com/quarkiverse/quarkus-roq/issues/1126

jtama (approved):
LGTM

Issue #1104, opened 2026-07-19
Auto-generate a per-page Markdown twin (`<page>.md`) for AI agents


rolfedh:
### Background

Roq already generates `llms.txt` and `llms-full.txt` (#793, shipped in #797). That is a good start, but the links in `llms.txt` point to rendered HTML pages, and there is no Markdown version of an individual page — for example `https://iamroq.dev/docs/getting-started/index.md` returns 404.

AI agents increasingly ask for Markdown directly. Claude Code sends `Accept: text/markdown` by default, and Vercel, Bun, Stripe, and Read the Docs now serve a Markdown copy of each page. Bun measured about 10x fewer tokens versus HTML. An Ahrefs study (June 2026) found 97% of llms.txt files get zero requests, so llms.txt alone is not the answer — per-page Markdown is what agents actually fetch.

### Proposal

For every generated page, also emit a Markdown twin at `<page-url>.md`. Static-file output, so it works on GitHub Pages with no server logic.

Content of each twin:
- The page body as clean Markdown — no site navigation or chrome.
- YAML front matter with at least `title` and `canonical_url` (Vercel and Stripe both do this), so an agent can cite the human page.
- A link back to the original source file.

Source handling:
- Markdown-authored pages: reuse the source.
- AsciiDoc-authored pages: resolve includes and attributes first (Quarkus guides use includes heavily), then convert. Two candidate routes — convert the reduced AsciiDoc source (asciidoctor-reducer plus a converter), or convert the rendered HTML (what `llms-full.txt` already approximates). Worth prototyping both on a few real guides to compare fidelity (admonitions, callouts, code tabs).

Discoverability:
- Point `llms.txt` links at the `.md` twins.
- Add `<link rel="alternate" type="text/markdown" href="....md">` to each HTML page's head.

Configuration:
- Opt-in per collection.
- Configurable URL pattern.
- Option to build `llms-full.txt` from the twins instead of regex-stripped HTML.

### Out of scope

True content negotiation (the same URL returns Markdown when the client sends `Accept: text/markdown`) needs an edge/CDN layer, which is a hosting decision for each site. A short recipe showing how a Cloudflare rule can map that header to the static `.md` twin would be enough for now.

### First consumer

quarkus.io guides — see quarkusio/quarkus#54921. If this lands, every Roq-built Quarkiverse site gets agent-ready docs for free.

I am happy to prototype this.

rolfedh:
[...]

Issue #133, opened 2024-10-05
should roq.java not do actual roq things?


maxandersen:
roq.java seem to be just serving content, like what `jbang httpd@jbangdev -d target/roq -p 8181` does for Java 8+
or since Java 18 there is `jwebserver -d target/roq -b 8181`

shouldn't we make `roq.java` do actual roq stuff?

ia3andy:
yeah I was thinking about adding the ability to generate but I am not sure how we could do it.

PR #110, opened 2024-09-24
Bump build.yml to JDK 21

gastaldi:
Trying to reproduce the error in https://github.com/quarkiverse/quarkus-roq/issues/108

PR #1105, opened 2026-07-20
Add markdown-twin plugin: Markdown twins of content pages for AI agents

rolfedh:
Publishes a Markdown twin of every content page for AI agents (#1104), produced from the source at build time. The twin lives at `<page-url>index.md` for a directory-style URL such as `/docs/basics/` and at `<page-url>.md` for a URL with a file name, the llms.txt convention for each case.

- Markdown pages: the verbatim Markdown source.
- AsciiDoc pages: parsed in-process with Yupiik's pure-Java `asciidoc-java` (already a Roq dependency; it resolves `include::` while parsing), then rendered to GitHub-Flavored Markdown by a new `MarkdownRenderer`. Source to source, never through HTML. No external tools. A cross-page `xref:page.adoc` points at the other page's twin, `../page/index.md`.

Each twin is emitted as a `RoqFrontMatterStaticFileBuildItem`, so Roq serves it at `/<page>/index.md` and writes it into the generated site with no runtime route. Keeping the twin inside the page's directory also keeps hosts that resolve an extensionless request to a sibling file (surge.sh does) from answering `/docs/basics` with the twin instead of the page. The existing `llmstxt: false` front-matter flag is the per-page opt-out. The site's `quarkus.asciidoc.attributes.*` plus `site-url`, `site-path`, `page-url`, `page-path`, `docname` and `sitegen` are passed to the parser, so attribute references resolve as they do for the rendered page. A page the parser rejects gets no twin and a warning; the build never fails because of a twin.

Tests: 26 unit tests on the renderer, the converter and the twin paths (deployment module, plain JUnit), and RoqBlogTest exercises the real pages, including the include-heavy docs pages, with no skipped test. See the sense-check comment for what is lost.

Follow-ups, not in this PR: point the llms.txt links at the twins (llms-full.txt and sitemap.xml stay as they are, see the discussion above), a `link rel="alternate" type="text/markdown"` tag per HTML page, front matter in the twins, filtering of theme and section pages, and the upstream Yupiik issues listed in the sense check.

Part of #1104; the discoverability items stay open on the issue.


holly-cummins:
It would be good to extend the testing for this to Asciidoc, since that's where I imagine many things can go wrong. I'd be tempted to add install steps for the prereq tools to CI, so we can get the coverage, pending a JVM-native solution. 

Does the new markdown get referenced in llms.txt or sitemap.xml?

rolfedh:
Thanks Holly. Agreed on the AsciiDoc coverage.

Could you and Andy steer me on one decision: go JVM-native in this PR, or keep it as the prototype with CI install steps for reducer and downdoc and do the JVM-native work in a follow-up? My proposal is JVM-native now.

Why:
- The flattening step is already solved. Roq ships Yupiik's pure-Java AsciiDoc parser (asciidoc-common runs it in a build step to parse headers), and it resolves includes while parsing, so reducer goes away.
- The only missing piece is a Markdown writer: a MarkdownRenderer visitor in the twin plugin's deployment module, roughly 600 lines that Roq would own. The ProcessBuilder code goes, and the AsciiDoc twin test runs unconditionally in CI.
- The trade-off, measured rather than guessed: with Yupiik 1.2.14, all 13 blog AsciiDoc pages parse and render (config tables included), and 247 of 279 quarkus.io guides do. The 32 failures share one cause, a strict callout check that rejects code blocks whose `<n>` markers do not match the callout list one to one (cdi-reference has three markers on one line). I would work around that in the converter and report it upstream to Yupiik. Neither corpus uses Asciidoctor extensions in AsciiDoc (no diagram blocks, tabs are CSS roles). A page Yupiik still rejects gets no twin and a warning, never a failed build.

On llms.txt and sitemap.xml: not yet. llms.txt, llms-full.txt, and sitemap.xml are Qute templates that iterate site.pages and the collections, and the twins are static files, so they are invisible to all three. My plan, open to your call:
- llms.txt: point its links at the .md twins in a small follow-up, falling back to the HTML URL for pages without a twin.
- llms-full.txt: leave as is. The June Ahrefs study of 137,000 domains found 97% of llms.txt files are never requested, and the little real traffic is coding agents such as Claude Code, which follow per-page links anyway. So the discoverability work that matters is the per-page .md twin plus a `link rel="alternate" type="text/markdown"` tag in each HTML page's head.
- sitemap.xml: leave alone, since the twins are alternates of existing pages rather than pages.

Thanks!

holly-cummins:
>     * The flattening step is already solved. Roq ships Yupiik's pure-Java AsciiDoc parser (asciidoc-common runs it in a build step to parse headers), and it resolves includes while parsing, so reducer goes away.

One caution here is how much we rely on Yupiik. I haven't thought about this very hard, so maybe this is a non-issue, but I know for Quarkus, Yupiik isn't powerful enough, and we need to render with the AsciidoctorJ parser. I wish I could think of a specific file or feature to use as a test example, but I can't.

So ideally this would delegate to the same asciidoc renderer used in the rest of the user's stack (but I haven't thought about how the current architecture relates to that, at all, so maybe it's a non-issue). I see you address this below.

We know yupiik are open to PRs, so if we can narrow down gaps, remediating in yupiik is a reasonable option. 

>     * The only missing piece is a Markdown writer: a MarkdownRenderer visitor in the twin plugin's deployment module, roughly 600 lines that Roq would own. The ProcessBuilder code goes, and the AsciiDoc twin test runs unconditionally in CI.
> 
>     * The trade-off, measured rather than guessed: with Yupiik 1.2.14, all 13 blog AsciiDoc pages parse and render (config tables included), and 247 of 279 quarkus.io guides do. The 32 failures share one cause, a strict callout check that rejects code blocks whose `<n>` markers do not match the callout list one to one (cdi-reference has three markers on one line). I would work around that in the converter and report it upstream to Yupiik. Neither corpus uses Asciidoctor extensions in AsciiDoc (no diagram blocks, tabs are CSS roles). A page Yupiik still rejects gets no twin and a warning, never a failed build.

We'd probably want to do a visual sense check here. Even if the parsing passes, is information lost? 

But based on that analysis, it seems promising enough that I think pure-Java is a good path to pursue. 

> On llms.txt and sitemap.xml: not yet. llms.txt, llms-full.txt, and sitemap.xml are Qute templates that iterate site.pages and the collections, and the twins are static files, so they are invisible to all three. My plan, open to your call:
> 
>     * llms.txt: point its links at the .md twins in a small follow-up, falling back to the HTML URL for pages without a twin.
> 
>     * llms-full.txt: leave as is. The June Ahrefs study of 137,000 domains found 97% of llms.txt files are never requested, and the little real traffic is coding agents such as Claude Code, which follow per-page links anyway. So the discoverability work that matters is the per-page .md twin plus a `link rel="alternate" type="text/markdown"` tag in each HTML page's head.

Ah, makes sense about the alternate! I agree about the plan for follow-up and leaving llms-full alone.

rolfedh:
Thanks Holly, that is what I needed. I will go ahead with the pure-Java path unless Andy sees it differently, and I am treating your two cautions as deliverables of this PR rather than caveats.

- Visual sense check: once the renderer exists, I will render the 13 blog pages and a sample of quarkus.io guides to Markdown, compare them side by side with the HTML, and post what is lost as a checklist here before asking for review. Parsing without error is only the first bar.
- Same renderer as the site: that needs a larger change than this PR. Roq renders AsciiDoc at request time inside a Qute section helper, the twins are produced at build time as static files, and neither engine has a Markdown backend, so the writer is new code either way. I picked Yupiik for the first cut because AsciidoctorJ in a build step means a JRuby start on every build, dev mode included. The conversion sits behind one small class, so an AsciidoctorJ-based mode can be added later if the sense check finds gaps Yupiik cannot close.
- Yupiik gaps: the callout strictness goes upstream as an issue with a repro, and a PR if they want one. Meanwhile the converter works around it so those pages still get a twin. I will keep a list of anything else the sense check turns up.

Next on my side: rebase onto main, then the renderer, converter, and tests. I will take the PR out of draft when the AsciiDoc twin test runs green in CI and the sense check is posted.

rolfedh:
Sense check, as promised. Method: after the renderer landed, I converted every AsciiDoc page of the Roq blog through the real build (RoqBlogTest, 16 tests green, none skipped), read the twins side by side with the rendered HTML for the docs pages and the AsciiDoc markup example page, and ran the converter directly over all 311 AsciiDoc files of the deployed quarkus.io guides as a larger sample.

What survives, checked page by page on the blog:
- Titles and section levels, including sections that come from includes.
- Source blocks with language and title; callout markers stay in the code and the callout list follows as a numbered list.
- Admonitions as GitHub alerts, including the ones written as open blocks.
- Tables as pipe tables, including the generated config-reference tables (header row, escaped pipes, cell content flattened to one line).
- Quotes with attribution, collapsible blocks as details elements, description lists, nested and checklist lists, images with the imagesdir prefix, cross-references pointing at the other page's twin.
- Attribute references defined by the page, its includes, or the site config resolve; Qute expressions and unknown references stay literal on purpose.

What is lost or degraded, and how:
- Block content inside a| table cells is flattened to a single line with br tags (config tables read fine; multi-paragraph cells lose their paragraph breaks).
- icon: macros are dropped, btn: and menu: become bold text, kbd: becomes kbd tags.
- A code block whose callout markers do not match the callout list one to one (three markers on one line, for example) is rejected by the parser; the converter retries with the callouts degraded to a plain numbered list, and the markers show as (1) instead of <1>. 30 of the 311 quarkus.io files take that path.
- 2 of 311 quarkus.io files get no twin at all: doc-create-tutorial.adoc (the parser resolves an include:: directive that sits inside a listing block, then trips on the example's callouts) and telemetry-micrometer.adoc (a NullPointerException inside the parser after the callout retry). Both are parser bugs; they go to Yupiik with the callout strictness and the '''-as-rule gap.
- Footnotes, %hardbreaks and counters are implemented but untested on real pages (the blog has none).

Numbers: blog 13 of 13 AsciiDoc pages, quarkus.io 309 of 311 files, 6.5 seconds for the whole quarkus.io set.

Upstream list for Yupiik (filed next): callout list strictness; ''' not recognized as a thematic break; include:: resolved inside listing blocks; NullPointerException in Reader.rewind on telemetry-micrometer.adoc; bare URLs keep trailing punctuation.

holly-cummins:
I guess we want to think about what the bar is for mergeability. That is, is *any* markdown better than none? Or, at the other end, do we need perfect fidelity? For a lot of changes I'd say going in the right direction and incremental improvement is sufficient. For this, I think we need to feel confident that LLMs will be 'better off' reading the markdown than the HTML. That is, they won't get confused or misled by what's in the markdown. People are more sensitive to this than LLMs, but we also want to be confident that they won't decide our docs are terrible and stop reading them/start recommending our competitors. It doesn't work exactly like that, I know, but LLMs don't like surprises. Well, technically, the LLMs find surprises expensive, so it's the people using them who don't like surprises. :) 

I think the blocks inside tables, icons, and keyboard is stuff we don't need to worry, since it's mostly visual. So it's looking pretty healthy. Thanks for doing that analysis, @rolfedh. Shall we raise the yupiik issues?

For the generation-faults, I guess we want a two-pronged approach. Well, maybe three. But we should start pushing fixes to Yupiik where we can, and where it's something in Quarkus that's breaking things, we should do fixes there. (Maybe in addition to making Yupiik tolerate more.) For example, fixes for https://github.com/quarkusio/quarkus/issues/56386 fixed a bunch of the callout issues, and I see that https://github.com/quarkusio/quarkus/pull/56449 and https://github.com/quarkusio/quarkus/pull/56222 fixed a bunch more.

We should also check that if markdown fails, we don't add the `alternative` tag to the document. That probably merits a test. With that guard, in a way we don't really care about the failures, apart from good intentions to come back and fix them in a follow-up.

rolfedh:
Thanks Holly. My read is that this is mergeable as it stands, with the alternate link and its test as the follow-up. Would you agree?

Why:

- The bar you set holds by construction. A twin is written only when the conversion succeeds, so a page that fails has no .md at its URL and nothing points to one. The conversion is source to source, so a twin never contains text the page does not. The remaining losses are the visual ones you named.
- The Yupiik issues are raised and mostly fixed: six issues and seven pull requests (yupiik/tools-maven-plugin#111 to #117). Romain reviewed all of them within the hour: yupiik/tools-maven-plugin#116 merged, yupiik/tools-maven-plugin#112, yupiik/tools-maven-plugin#113, yupiik/tools-maven-plugin#114 and yupiik/tools-maven-plugin#115 approved, yupiik/tools-maven-plugin#111 and yupiik/tools-maven-plugin#117 in discussion (callout positions carried in the model rather than re-parsed in the renderer, and failure by default with an opt-in lenient attribute; I agree with both). When 1.2.17 ships, the converter's callout workaround comes out.
- You are right about the source side, and it improves my numbers. The sense-check corpus was a July snapshot, before quarkusio/quarkus#56222, quarkusio/quarkus#56425 and quarkusio/quarkus#56449. On today's main (327 files) the released parser fails 20, all on the callout check: ten with several markers on one line, four with + continuations in the callout list, the rest shapes the fixes also cover. The patched parser parses all 327.
- The alternate tag is not in this PR. It is on the follow-up list, and when it lands it will be driven by the set of twins actually generated, with the test you describe, so a failed page can never advertise one. Happy to pull it into this PR instead if you would rather have it here.

holly-cummins:
Thanks Rolfe, I agree on the merging. That's great news on the yupiik fixes. Do you mind raising a quarkus defect for the remaining callout issues, since you have them in front of you?

rolfedh:
Done: quarkusio/quarkus#56509 lists the seven blocks in six guides (a marker without an entry in four, a block without a list in two, an entry cut off from its list by a paragraph in one), with line links on today's main. The eighth block a strict parser trips on, in doc-create-tutorial.adoc, sets subs="-callouts" on purpose, so that one went to Yupiik as yupiik/tools-maven-plugin#118 instead. I also qualified the cross-repository numbers in my earlier comment so they point at Yupiik rather than here.

ia3andy:
it looks cool :)

maxandersen:
i'm not following how generating a twin for parts of the site helps agents - how will agents "discover" these ?

fwiw, for jbang.dev I simply enabled https://www.searchenginejournal.com/cloudflares-new-markdown-for-ai-bots-what-you-need-to-know/567339/ which makes it so markdown is served for a request when accept markdown is present in headers from agents.

for tamboui.dev (not using cloudflare) i generate https://tamboui.dev/llms.txt that passes key files through cloudflare md generator.

Either of above would let agents actually reach/see the markdown without additional work / cost on the site?

rolfedh:
> i'm not following how generating a twin for parts of the site helps agents - how will agents "discover" these ?

Fair question. #1105 on its own ships no discovery; that is in two follow-ups agreed with Holly above.

- #1224 (draft on top of this one): every HTML page whose twin exists gets `<link rel="alternate" type="text/markdown" href="<page-url>index.md">`, the link relation the [llms.txt spec, v2 of 2026-08-10](https://llmstxt.org/) defines for this, and what [Vercel's](https://vercel.com/docs/agent-resources/markdown-access) and [Cloudflare's](https://developers.cloudflare.com/fundamentals/reference/markdown-for-agents/) docs pages emit. The [preview](https://quarkiverse-roq-blog-1224-preview.surge.sh/docs/basics/) carries it and serves the twin as `text/markdown`.
- #1219: Roq's existing llms.txt gets links to the twins. The spec: "The links in an llms.txt file should therefore point to LLM-friendly content, such as the markdown versions of pages described above."
- The URL itself: the spec's convention for a URL without a file name is `<page-url>index.md`, which is where the twin now lives.

Which agents follow the alternate link or read llms.txt is not something I have measured. [Checkly's February 2026 survey](https://www.checklyhq.com/blog/state-of-ai-agent-content-negotation/) covers a different mechanism, the `Accept: text/markdown` header, which GitHub Pages ignores.

Edited 2026-09-11: dropped the Stripe example (Stripe also ships llms.txt and content negotiation), corrected the URL convention for directory-style URLs, and stopped citing the Checkly survey as evidence of discovery. Sources linked inline.

rolfedh:
> fwiw, for jbang.dev I simply enabled https://www.searchenginejournal.com/cloudflares-new-markdown-for-ai-bots-what-you-need-to-know/567339/ which makes it so markdown is served for a request when accept markdown is present in headers from agents.

It works well for jbang.dev, and quarkus.io turns out to be a partial exception to what I first wrote here. quarkus.io does answer `server: cloudflare`, with GitHub Pages and Fastly behind it. iamroq.dev, docs.quarkiverse.io and tamboui.dev answer `server: GitHub.com` with no Cloudflare layer. All four still return HTML to `Accept: text/markdown` today, and quarkus.io sets `vary: Accept-Encoding` with no `Accept` in the Vary, so no content negotiation is happening anywhere right now. The difference is that on quarkus.io there is a layer where it could be configured, and on the other three there is not.

The feature itself needs the site behind Cloudflare on a Pro, Business or Enterprise plan ([docs](https://developers.cloudflare.com/fundamentals/reference/markdown-for-agents/)), and it converts the origin HTML at the edge, so the page chrome comes along: the Markdown it returns for the [jbang.dev docs index](https://www.jbang.dev/documentation/jbang-all/latest/index.html) starts with 21 lines of component menu before the title on line 22.

The two are complementary rather than exclusive, and for quarkus.io that is not hypothetical. A site behind Cloudflare that has twins can serve them instead of converting, with a URL rewrite rule keyed on the Accept header; [Transform Rules](https://developers.cloudflare.com/rules/transform/) exist on every plan, ten on Free, so this path does not carry the Pro-plan requirement and it serves clean twins rather than edge-converted chrome. Cloudflare's own docs site behaves this way already: `Accept: text/markdown` on a page returns the same bytes as that page's `index.md`. I have drafted the two rules and checked them against the plugin's twin-path logic, but I have not run them on a real zone.

Edited 2026-09-11: removed "the only form that works on that hosting" (Roq already serves llms.txt and llms-full.txt as static files there) and marked the rewrite-rule idea as untested. Sources linked inline.
Edited again 2026-09-11: corrected the claim that quarkus.io is not on Cloudflare. It is; the other three sites are not.

rolfedh:
> for tamboui.dev (not using cloudflare) i generate https://tamboui.dev/llms.txt that passes key files through cloudflare md generator.

That works, and it asks nothing of the site. Two costs, though. All six links in that llms.txt go through markdown.new, a third-party service by Emre Elbeyoglu of Popupsmart built on Cloudflare Workers AI and Browser Rendering, limited to 500 requests per day per IP, with terms that say "as is" and "as available" ([FAQ](https://markdown.new/), [terms](https://markdown.new/terms)). And because it converts the rendered HTML, the page chrome comes along: for [quarkus.io/guides/getting-started](https://markdown.new/https://quarkus.io/guides/getting-started) the title lands on line 29 after the version menu, the footer takes lines 418 to 481 of 481, and the six admonitions become two-column tables with an empty icon cell.

A twin is built from the AsciiDoc or Markdown source, once, at build time, for every page authored in Markdown or AsciiDoc that converts, and it costs nothing when fetched: a static file, no service in the path.

Edited 2026-09-11: corrected the description of markdown.new (it has a named operator), the admonition count (six, text kept, not twelve empty tables), and the coverage claim (HTML-authored pages get no twin); dropped the timing figure.

rolfedh:
> Either of above would let agents actually reach/see the markdown without additional work / cost on the site?

Each option puts its work somewhere.

- Cloudflare's feature: the site goes behind Cloudflare, which on the Free and Pro plans means Cloudflare becomes the DNS provider ([CNAME setup is Business and Enterprise only](https://developers.cloudflare.com/dns/zone-setups/partial-setup/)), on a paid plan or on a free one through [Project Alexandria](https://www.cloudflare.com/lp/project-alexandria/) for open source projects. For quarkus.io that is a project decision, and every Roq site would make it separately.
- llms.txt through markdown.new: nothing on the site. The work moves to each fetch, agent to markdown.new to the site, against a shared limit of 500 requests per day per IP and no availability commitment.
- The twin: build time and static files, no DNS change, no plan, no service in the path. The site adds the plugin, and the per-page opt-out is the `llmstxt: false` flag it already has.

Edited 2026-09-11: dropped the timing figure (the measurement was 309 of 311 files in a standalone converter run) and "change its hosting" (the origin stays on GitHub Pages; DNS and proxying change). Sources linked inline.

rolfedh:
@maxandersen you were right to push on this. I went back over my four comments above against the sources and corrected them in place; the edit history is under each "edited" mark, and sources are now linked inline. What changed:

- markdown.new is not operator-less. It is a third-party service by Emre Elbeyoglu of Popupsmart, built on Cloudflare Workers AI and Browser Rendering, limited to 500 requests per day per IP, provided "as is".
- Stripe does not rely on the `.md` URL alone. It also ships llms.txt and answers `Accept: text/markdown`.
- The Checkly survey shows which agents send `Accept: text/markdown`. It says nothing about agents following `rel="alternate"` or reading llms.txt, and I had used it as if it did.
- For a directory-style URL such as `/docs/basics/`, the llms.txt spec's guessable name is `/docs/basics/index.md`, not `/docs/basics.md`; Cloudflare's docs, which have the same URL shape, do exactly that. I have moved the twin there in #1105 (pushed today), which also removes the no-slash collision noted on #1224.
- The quarkus.io getting-started page has 6 admonitions, not 12, and markdown.new keeps their text, in a table with an empty icon column.
- "6.5 seconds for all 311 guides" was 309 of 311 in a standalone converter run, and the twin skips pages authored in HTML.
- "A Cloudflare rule can serve the .md instead of converting" was untested. What is documented: Transform Rules exist on every plan, ten on Free, and can match on the Accept header, so a site that has twins can do content negotiation on Cloudflare without the paid feature.

Short answer to your question: #1105 on its own gives agents nothing to discover. Discovery comes from #1219, llms.txt links that point at the twins, which is the path with evidence of agent use today, and from #1224, the `rel="alternate" type="text/markdown"` tag that the llms.txt v2 spec added in August and that Vercel ships; I have no measurement yet of which agents follow it. Both of your alternatives compose with the twins: a Cloudflare-fronted site can serve the twin instead of converting HTML, and llms.txt can point at twins instead of markdown.new.

rolfedh:
Scoping follows separately in #1234: `enabled` plus `include`/`exclude` globs, so a site can keep the plugin installed without twinning everything. Nothing to change here.

rolfedh:
A question on where the Markdown renderer should live. Two projects want the same conversion: this plugin, and the Quarkus documentation RAG pipeline (quarkusio/quarkus-documentation-rag#12), which cannot depend on Roq. I have asked the Yupiik parser project whether they would take the renderer as `GithubFlavoredMarkdownRenderer` (yupiik/tools-maven-plugin#130), and Romain has said yes in principle.

My suggestion is to leave the renderer in this pull request as it is, and once a Yupiik release ships it, open a follow-up here that deletes the copy and bumps the version. @holly-cummins, on relying on Yupiik: the renderer is already a `Visitor` over Yupiik's model, so the dependency is the same wherever the class sits.

@ia3andy, @holly-cummins, does that sound right to you, or would you rather keep the renderer in Roq?

rolfedh:
I pushed one small fix. When a code block had callouts, the renderer turned any code line ending in a number in parentheses into a callout marker: 19 code lines in 10 quarkus.io guides changed, for example `.statusCode(200)` became `.statusCode<200>` in getting-started. It now restores `(n)` only after a space or a comment prefix such as `//`, and only when `n` is one of that block's callouts, close to what the Yupiik HTML renderer does. I found it while comparing this renderer with the Yupiik port on all 279 guides. Nothing else changed.

rolfedh:
The Markdown renderer from this pull request is now proposed to the Yupiik parser as yupiik/tools-maven-plugin#133, with the fixes found since then. Nothing changes here; Roq can switch to it after a Yupiik release.

rolfedh:
@holly-cummins @ia3andy, a correction: this pull request still has bugs that put wrong text in a twin. I propose to fix them in one small commit before this merges, or in a follow-up if you would rather merge first. Which do you prefer?

For example, `pass:q[*raw*]` comes out as `q`, and a code block that contains four backticks turns the rest of the page into code.

This means my comment of 8 September was wrong. I wrote that "a twin never contains text the page does not" and that "the remaining losses are the visual ones you named". The review of the Yupiik version of this renderer (yupiik/tools-maven-plugin#133) found these bugs in the code it copied from here. I should have said so when I opened #133.

Other bugs make a twin harder to use, and they wait for the switch to the Yupiik renderer: paragraphs inside an admonition are glued together, and links to `[[id]]` anchors lead nowhere.

rolfedh:
I decided to fix the wrong-text bugs from my comment above before this merges, so I marked this pull request as a draft. I will write here when the fix is pushed.

rolfedh:
I pushed 8a796f14, which fixes the bugs that moved, lost or added text in a twin, so this is ready for review again.

The commit fixes list text moved above a code block, a code fence closed by backticks in the code, lost `kbd` keys, `pass:q[]` printed as `q`, stray `**` around `menu:[]`, a `[[id]]` read as a fence language, footnotes defined twice or showing their id, and `indexterm` text shown. Each fix has a test. The commit also keeps a list item's first paragraph on its marker line, which removes most empty list markers.

Two text losses remain in the Yupiik parser, and Roq will get their fixes by updating `asciidoc-java`, with no change to this plugin. Text after a comma in `footnote:[...]` or in a link text is dropped, and so is a word written right before a macro, as in `name.footnote:[...]`. A page rendered with the `asciidoc` plugin loses the same text, because that plugin uses the same parser. Both are in [yupiik/tools-maven-plugin#142](https://github.com/yupiik/tools-maven-plugin/issues/142), and I plan to fix them after the higher-priority gaps in that list.

Glued paragraphs and dead `[[id]]` links wait for the switch to the Yupiik renderer ([yupiik/tools-maven-plugin#133](https://github.com/yupiik/tools-maven-plugin/pull/133)), which needs parser changes that are still in review. The switch is a follow-up pull request that updates `asciidoc-java` and deletes this renderer, after a Yupiik release contains those changes. I do not have a release date yet.

rolfedh:
Corpus re-check on today's quarkusio/quarkus main, now that the six callout fixes are merged and #56509 is closed.

Short version: the callout work is done. What is left is a Yupiik release, not a source problem in the guides and not a converter problem.

Numbers, over all 328 .adoc files of quarkusio/quarkus main at 0cb7eba99a3, 288 of them top-level guides:

- asciidoc-java 1.2.16, the version this PR pins: 326 files get a twin, 12 of them through the degraded callout path, and 2 get none.
- A local build of 1.2.17-SNAPSHOT, carrying the nine approved Yupiik fixes: all 328 convert cleanly, nothing degraded and nothing skipped.

The degraded count was 30 on 2026-09-08, and none of the six guides the upstream fixes touched is still on the list, so those fixes did what they were meant to do. The two files with no twin are the same two as then, doc-create-tutorial.adoc and telemetry-micrometer.adoc, both parser defects already filed as yupiik/tools-maven-plugin#109 and #110. All 14 of those files convert on 1.2.17-SNAPSHOT.

I also found one thing in the converter itself, and I have a fix ready to push to this branch. convert() caught IOException and RuntimeException, so a StackOverflowError escaped it and failed the build, although the javadoc promises that a conversion problem only costs the page its twin. Four lines reach it on 1.2.16:

```asciidoc
= Self reference
:a: {a}

See {a} here.
```

The parser evaluates attribute values lazily, so a value naming its own attribute recurses at render time. That is yupiik/tools-maven-plugin#122, fixed there, but the guard belongs here too: 87 of the 328 guides hit it before the docs build substitutes the Maven placeholder in _attributes.adoc. The commit widens the catch to include StackOverflowError, corrects the javadoc, and adds a regression test asserting that the converter does not throw, so it holds whichever parser version the build pins.

Method, if you want to repeat it: the converter from this branch, run over every file with the Maven placeholders substituted the way the docs build substitutes them before the site receives the sources.

rolfedh:
@maxandersen coming back to your question, in short form. I don't see the two as competing.

- A site on a Cloudflare Pro, Business or Enterprise plan can turn on Markdown for Agents and switch the twins off: `llmstxt: false` per page today, and site-wide the `enabled` and include/exclude options proposed in [#1234](https://github.com/quarkiverse/quarkus-roq/pull/1234).
- A Roq site without such a plan gets the twin, the option that needs no service in the path.
- Cloudflare's own docs site does both: it answers the Accept header and also publishes an `index.md` twin with the alternate link.
- One thing worth weighing between the two is latency. Cloudflare converts the HTML on the fly; a twin is a static file the host serves as is, with its cost paid once at build time. Whether that difference changes how AI crawlers treat a site, I don't know, but it is a question worth asking.

On how an agent finds a twin, there are three ways to reach it, and I have not found a measurement of which agents follow them:

- The address the llms.txt convention makes guessable, `<page-url>index.md`.
- The `<link rel="alternate" type="text/markdown">` that [#1224](https://github.com/quarkiverse/quarkus-roq/pull/1224) adds to every page that has a twin.
- The llms.txt links, which [#1219](https://github.com/quarkiverse/quarkus-roq/issues/1219) is the follow-up to, point at the twins.

The address and the link are what [revision 2 of the llms.txt proposal](https://llmstxt.org/), published 10 August 2026, recommends.

maxandersen:
Rolf, I seriously can't grok that agent comment. Can you please write it as a human. Thanks.

rolfedh:
> Rolf, I seriously can't grok that agent comment. Can you please write it as a human. Thanks.

@maxandersen sure, sorry about that. Let me try again, against what you actually asked.

The short version: the question for Roq is whether this is worth offering to every site built with Roq, not just jbang.dev and quarkus.io, which happen to sit behind Cloudflare. Roq's own site, iamroq.dev, is on plain GitHub Pages and can't flip that switch. A twin built from the source is a cleaner file than any conversion of the rendered page, and Cloudflare's own docs serve a pre-generated file the same way. The renderer, which is most of the plugin's code, is moving to the Yupiik parser project, so what stays in Roq is small. What nobody has measured yet is whether agents follow the links to the twins, and I'd rather say that than claim it.

> how will agents "discover" these ?

Three ways from the site's side. The twin sits next to the page, at `<page-url>/index.md` (for example `/docs/basics/index.md`), so an agent that knows the llms.txt convention can guess the address. Each page gets a `<link rel="alternate" type="text/markdown">` pointing at its twin ([#1224](https://github.com/quarkiverse/quarkus-roq/pull/1224)). And llms.txt will list the twins instead of the HTML pages ([#1219](https://github.com/quarkiverse/quarkus-roq/issues/1219)). That's what version 2 of Jeremy Howard's [llms.txt proposal](https://llmstxt.org/), revised on 10 August 2026, recommends, and it's what Cloudflare's and Vercel's docs ship. As far as I know, nobody has measured whether agents actually follow those links.

The fourth route, which you mentioned earlier, is different in kind. Some agents send an `Accept: text/markdown` request header, which asks the server for Markdown instead of HTML at the same URL. That is the only route with measured demand today: [Checkly saw](https://www.checklyhq.com/blog/state-of-ai-agent-content-negotation/) Claude Code, Cursor and OpenCode sending it in February, and [by June](https://acceptmarkdown.com/status) GitHub Copilot did too. But answering it takes a server that looks at the header, and GitHub Pages doesn't, so a GitHub Pages site needs Cloudflare or a similar layer in front to serve anything on that route, which is what you did for jbang.dev.

> Either of above would let agents actually reach/see the markdown

Reach it, yes. See the same thing, no. I ran the getting-started guide through the same converter you use for tamboui.dev ([the result](https://markdown.new/https://quarkus.io/guides/getting-started)), and built that guide from its AsciiDoc source with the twin renderer:

| | converted from HTML | built from source |
|---|---|---|
| title | line 29, after the version menu and the guide categories | line 1 |
| admonitions | tables with an empty icon cell | `> [!TIP]` alerts |
| footer | Commonhaus link, copyright, trademark policy | none |

[Cloudflare's own docs](https://developers.cloudflare.com/style-guide/how-we-docs/ai-consumability/) do what I'm proposing, and they say why. Every page has a pre-generated `index.md` twin, linked from the page with `<link rel="alternate" type="text/markdown">`. A request with the `Accept: text/markdown` header gets that same file back, byte for byte, without the `x-markdown-tokens` header that their on-the-fly conversion adds.

Their reason is tokens: they count their Workers Get Started page at 15,229 tokens as HTML and 2,110 as Markdown. Their file is made from their rendered HTML, so it still carries a skip link and a toolbar line, but the pattern is generate once, then answer both the `/index.md` address and the `Accept: text/markdown` header with that one file. On quarkus.io that would be a rewrite rule mapping the header to the twin, which exists on every plan. I haven't tried that yet.

On llms.txt through a converter, the trouble is that crawlers rarely fetch llms.txt at all. [Ahrefs found](https://ahrefs.com/blog/llmstxt-study/) in May 2026 that 97% of the files got no request, so that route has the same discovery problem as the twins, plus a third party with a daily request limit.

> without additional work / cost on the site?

You're right that your path is zero site work, and quarkus.io sits behind Cloudflare just like jbang.dev. So Markdown for Agents there would be a switch on a paid plan. The plugin costs a site nothing unless it opts in: if you don't add it, nothing changes, and with [#1234](https://github.com/quarkiverse/quarkus-roq/pull/1234) you can twin part of a site or none of it, which also keeps the build cost out of dev mode. What it does cost is a module in Roq. That's real, but it shrinks soon: the renderer and its tests are about three quarters of the module, and they are moving to the Yupiik parser project ([yupiik/tools-maven-plugin#133](https://github.com/yupiik/tools-maven-plugin/pull/133), approved today), so Roq keeps only the build step and the path rules. Happy to dig into any of this further if it would help.

holly-cummins:
@maxandersen, see also https://github.com/quarkusio/quarkus-documentation-rag/issues/12. Our `documentation-rag` project needs to solve the same problem (converting our docs to markdown). So this is capability we need, independent of web-serving. @phillip-kruger had implemented something in his project, but after evaluating @rolfedh's version, concluded it produced more accurate markdown than what he'd done. So going forward, the work here will also be supporting the documentation RAG work.

Issue #1083, opened 2026-07-13
Allow arbitrary user-defined properties that can be varied depending on the config profile


holly-cummins:
This is a pattern that's used in Jekyll, and could be useful here as well. For example, https://github.com/quarkusio/quarkusio.github.io/blob/main/_config_dev.yml sets a different value of `search.hosts` for dev mode. 

At the moment, any `%profile.some.prop` config in application.properties has to be pre-defined in a `@ConfigMapping`. So profiles work well for 'official' properties, but not so well for other properties. I can see the following possible fixes:

- * User `@ConfigMapping`s*. Users could create a Java interface in `src/java` with a `@io.smallrye.config.ConfigMapping(prefix =   "my.prefix")` annotation, and once that's done, values could be set in `application.properties`, or with environment variables, and there would be full profile support. That pre-declaration in Java isn't great, and it gets worse, because the config wouldn't be a named bean, so access would be  something like `{cdi:siteConfig.get().search().host()}`. Roq could make this a bit better by auto-wiring discovered config interfaces into `@TemplateGlobal` or `@TemplateData`.
- *Catch-all map`. Properties which have a map type can have arbitrary keys. If Roq provided a standard config property,  templates could do `site.data.getString('search-host')`. This would  get profiles for free, but the data access pattern is pretty clunky, there's no hint of type safety or compile-time checking, and nested properties/keys would be impossible. 
- *Profile-aware data files`. This would mirror the Jekyll `_config_dev.yml` convention with a `_data/siteConfig_dev.yml`. That preserves the existing `cdi:` access pattern unchanged, and leverages Quarkus's existing profile infrastructure for activation. Roq would filter the files in `_data`  to use the right profile where there are multiple. The implementation would be in `RoqDataReaderProcessor.scanDataFiles`. When grouping scanned files by name, it detect the `_<profile>` suffix, check it against active profiles, and deep-merge before emitting the `RoqDataBuildItem`.

jtama:
So what you mean is that we would leverage Roq Data to do configuration, and filter files base on quarkus launch mode ?

holly-cummins:
> So what you mean is that we would leverage Roq Data to do configuration, and filter files base on quarkus launch mode ?

That's right. I don't love config being in the data rather than the application.properties, but I think it's the only way to enable arbitrary 'config' without writing Java code to pre-declare it. And it does kind of make sense that varying data based on launch mode might be useful.

ia3andy:
[...]

PR #1108, opened 2026-07-20
Conversion: Updates to config and qute converters to handle arbitary config and other bits

### LINKED ISSUE #1069: Conversion: Lighthouse SEO  need fixes after converting Jekyll to Roq
I'm not sure if this is a converter issue or a Roq issue, but since they live in the same repo, raising this here while I figure which code to blame. :)

The lighthouse reports SEO ranking for quarkus.io dropped from [92](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1782923375367-10880.report.html) to [75](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1783347017042-92460.report.html) after conversion.

Before:

Image

After:

Image

There are no further details on `Document does not have a valid rel=canonical Is not an absolute URL (/)`.

The hreflang seems to be something about translation:

```
head > link
<link rel="alternate" hreflang="en" href="NOT_FOUND" />
Relative href value
head > link
<link rel="alternate" hreflang="pt-br" href="NOT_FOUND" />
Relative href value
head > link
<link rel="alternate" hreflang="es" href="NOT_FOUND" />
Relative href value
head > link
<link rel="alternate" hreflang="zh" href="NOT_FOUND" />
Relative href value
```

holly-cummins:
This is a whopper, sorry. :( 

The headline feature here is conversion-time support for what's proposed in the discussion of in https://github.com/quarkiverse/quarkus-roq/issues/1083. The converter looks in the config files, and if any have _profile variants, it knows that config cannot be handled by something in data. So instead it converts that config to live in application.properties, and updates all the accesses to that data to use the different format. It also creates a config mapping and config bean to enable access from templates.

It also fixes https://github.com/quarkiverse/quarkus-roq/issues/1069, an SEO score regression. The cause was that the converter was creating lots of `concat()` in its qute when it saw liquid `|append`, but that method doesn't exist. I've gone back to `append`, but now it exists.

Finally, there are various bit of improvement: 
  - Better handling of mutable variables (variables used after their scope ends, see https://github.com/quarkiverse/quarkus-roq/pull/1101
  -  Smarter detection of variables that need MutableMap vs {#let}
  - New replaceInQuteBlocks() method to safely replace patterns only inside Qute expressions
  - Improved site.url conversion to allow method calls like .append() while blocking property access - this allows site.url.append(path) while still converting bare site.url to site.url.root.url, and fixes an SEO issue
  - Better handling of site.tags with .asStrings for flexibility. The asStrings method handles both yaml or a comma-separated string, but .orEmpty handled lists only

### DIFF
--- a/migration/src/test/java/io/quarkus/tools/LiquidToQuteConverterTest.java
+++ b/migration/src/test/java/io/quarkus/tools/LiquidToQuteConverterTest.java
@@ -1,6 +1,11 @@
 package io.quarkus.tools;
 
-import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.List;
 
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
@@ -13,6 +18,7 @@ class LiquidToQuteConverterTest {
     @BeforeEach
     void setUp() {
         converter = new LiquidToQuteConverter();
+        converter.setConfigMappingSections(List.of("arbitrary"));
     }
 
     private void assertConverts(String input, String expected, String message) {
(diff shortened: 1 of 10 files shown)


migration/src/test/java/io/quarkus/tools/LiquidToQuteConverterTest.java
```diff
@@ -430,15 +436,15 @@ void testTernaryWithoutMethodCall() {
     @Test
     void testAppendFilter() {
         String input = "{{\"hello\" | append: \" world\"}}";
-        String expected = "{=\"hello\".concat(\" world\").raw}";
-        assertConverts(input, expected, "Append filter should convert to .concat()");
+        String expected = "{=\"hello\".append(\" world\").raw}";
```
jtama, line 438: That escaping inside the cute content is weird.

holly-cummins: I think it's because it's in a string, so it's actually representing 

```
{="hello".append(" world").raw}
```

I could do a multiline string which would avoid the escaping (but also be weird, because there's only one line). But more readable, but also weirder, I think?

```
String expected = """
      {="hello".append(" world").raw}""";
 ```

jtama (approved):
Good to merge once conflicts are resolved.

PR #1109, opened 2026-07-20
Add converter to convert adoc links to xref

holly-cummins:
In adoc, it's preferred to use `xref:` instead of `link:` for internal cross-references. In Jekyll, `link:` still works, but in Roq, it does not. This converter is a lot of code for a small problem, but it finds the problem `link:` elements and converts them.


ia3andy (approved):
LGTM

PR #1110, opened 2026-07-20
Handle references to files with parentheses in the name in conversion to avoid 404

holly-cummins:
Roq's static file handler silently renames files with parentheses (e.g. `{@code content(1)}` becomes `{@code content-1})` during the build, but does not update HTML references, so images 404 at runtime.

This class fixes the source files before the build: it renames files to replace parentheses with dashes and updates all {@code .html} files in the same tree.

I wasn't totally sure if this is something that would be useful outside conversion, since any project could have this issue. But it seems too heavy to add elsewhere.

Btw, I know all these `*Command`s are getting excessive. Once everything is in main I will rewrite `roq-it-jekyll` to use/be a java orchestrator, but I don't want to do that refactoring until I have a working version in main. My branch and file and PR management is getting too complex as it is without adding a refactor on top.

### DIFF
--- a/migration/src/main/java/io/quarkus/tools/ParenthesisFileRenamer.java
+++ b/migration/src/main/java/io/quarkus/tools/ParenthesisFileRenamer.java
@@ -1,0 +1,76 @@
+package io.quarkus.tools;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+/**
+ * Renames files with parentheses under a directory and updates HTML references.
+ * <p>
+ * Roq's static file handler silently renames files with parentheses
+ * (e.g. {@code content(1)} becomes {@code content-1}) during the build, but does not
+ * update HTML references, so images 404 at runtime.
+ * <p>
+ * This class fixes the source files before the build: it renames files to replace
+ * parentheses with dashes and updates all {@code .html} files in the same tree.
+ */
+public class ParenthesisFileRenamer {
+
+    private static final Pattern PAREN_REFERENCE = Pattern.compile("([a-zA-Z0-9_-]+)\\((\\d+)\\)");
+
+    public record Result(int filesRenamed, int htmlFilesUpdated) {
+    }
+
+    public Result rename(Path directory) throws IOException {
+        if (!Files.isDirectory(directory)) {
+            return new Result(0, 0);
+        }
+
+        // Pass 1: find and rename files with parentheses
+        List<Path> htmlFiles = new ArrayList<>();
+        Map<Path, String> renames = new HashMap<>();
+
+        Files.walkFileTree(directory, new SimpleFileVisitor<>() {
+            @Override
+            public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
+                String name = file.getFileName().toString();
+                if (name.endsWith(".html")) {
+                    htmlFiles.add(file);
+                }
+                if (name.contains("(")) {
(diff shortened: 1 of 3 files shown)


migration/src/main/java/io/quarkus/tools/ParenthesisFileRenamer.java
```diff
@@ -0,0 +1,76 @@
...
+ * Roq's static file handler silently renames files with parentheses
+ * (e.g. {@code content(1)} becomes {@code content-1}) during the build, but does not
+ * update HTML references, so images 404 at runtime.
+ * <p>
+ * This class fixes the source files before the build: it renames files to replace
+ * parentheses with dashes and updates all {@code .html} files in the same tree.
+ */
+public class ParenthesisFileRenamer {
+
+    private static final Pattern PAREN_REFERENCE = Pattern.compile("([a-zA-Z0-9_-]+)\\((\\d+)\\)");
```
jtama, line 28: That regexp seems to capture many things. Isn't it **_too_** open ?

ia3andy: Yes review by claude told me the same:
 Three issues to address:

  1. Regex/rename mismatch: The file rename handles any parentheses (foo(bar) → foo-bar) but the HTML regex only matches word(digits). Files with non-numeric parens get renamed without their HTML references being updated, causing
  the very 404s this tool aims to fix.
  2. Rename collision risk: If content-1 already exists alongside content(1), Files.move may silently overwrite or throw depending on the platform. A pre-check or explicit handling would be safer.
  3. HTML replacement too broad: The regex matches parenthesized numbers anywhere in HTML, not just file paths. Text like section(3) in prose would become section-3. The replacement should be scoped to src/href attributes.

cescoffier:
Where are we on this?

ia3andy:
we added reviews with @jtama, maybe @holly-cummins haven't seen them?

holly-cummins:
> we added reviews with @jtama, maybe @holly-cummins haven't seen them?

Oh you're quite right, I didn't. Thanks for the pointer!

migration/src/main/java/io/quarkus/tools/ParenthesisFileRenamer.java
holly-cummins: #2 is a good point, and of course it affects the current Roq implementation as well (I raised https://github.com/quarkiverse/quarkus-roq/issues/1218). I guess here it's slightly worse because files could be overwritten, but I kind of wonder if the same solution shouldn't be done here and for #1218. 

For #1 and #3, those are good spots. I've checked and Roq itself will turn something like  `(hi(1).html` into `-hi-1.html`, so I think the rename is correct and the regexp needs to be broadened. And then it also needs to be scoped to href, as per #3.

holly-cummins: Actually, it's probably easy enough to throw, and worry about #1218 later. So I've done that, and also made the match between href and filename stricter, and scopred down the matching to only within href and src blocks.

holly-cummins: It should be a lot safer now, hopefully.

ia3andy:
@holly-cummins 

Finding 1 (High) — HTML parentheses cause NoSuchFileException
When a rendered filename contains parentheses (e.g. from a title like "Foo (bar)"), the rename path becomes stale and the file operation throws NoSuchFileException. The path is computed before the rename but used after.

Finding 2 (High) — Two sources mapping to the same target pass the pre-check
The conflict pre-check has a gap: if two different source files both rename to the same target, only the first is caught. The second overwrites silently because the check uses the wrong key direction.

Finding 3 (Medium) — No rollback on partial failure
If a batch rename partially succeeds before an error, already-renamed files are left in the renamed state with no cleanup. This can leave the workspace in an inconsistent state.

Finding 5 (Low) — !renames.containsKey(target) is dead code
Rename targets never contain parentheses (they're derived from slugs), so this branch of the conflict check never fires. It makes the pre-check look more complete than it is, reinforcing finding #2.

holly-cummins:
Next set of comments addressed, thanks!

ia3andy (approved):

PR #1112, opened 2026-07-20
Jekyll conversion: Convert redirects to aliases

holly-cummins:
Problem: Jekyll sites (like quarkusio.github.io) often have a _redirects/ directory containing standalone redirect files that map old URLs to new ones. During Jekyll → Roq migration, these redirects would be lost if the directory is simply deleted.

  Example from quarkusio.github.io:
  _redirects/guides/hibernate-search-elasticsearch/index.md
  This file has frontmatter:
  ---
  aliases:
    - /guides/hibernate-search-orm-elasticsearch
  --- 
  
  This means: "Redirect /guides/hibernate-search-elasticsearch →  /guides/hibernate-search-orm-elasticsearch"

  `RedirectMergerCommand` merges redirects into their target files as `aliases:` before deleting the redirects directory:

  1. Finds redirect files in content/redirects/
  2. Extracts source path from redirect file location:  /guides/hibernate-search-elasticsearch
  3. Extracts target URL from the redirect's frontmatter: /guides/hibernate-search-orm-elasticsearch
  4. Finds the target file:  content/guides/hibernate-search-orm-elasticsearch.adoc
  5. Adds the source path as an alias to the target file's frontmatter
  6. Deletes the redirect file (now redundant)

This was quite tricky to get right, and there were several edge cases that messed up early flows:
  
  - External URLs: Keeps redirect files for external redirects (e.g.,  https://example.com)
  - Circular redirects: Deletes useless redirects where source = target
  - Missing targets: Keeps redirect files when target file can't be found
  - Conflicts: Deletes redirect files that conflict with actual content at the  same path
  - Duplicate aliases: Won't add the same alias twice

### DIFF
--- a/migration/src/main/java/io/quarkus/tools/migration/command/RedirectMergerCommand.java
+++ b/migration/src/main/java/io/quarkus/tools/migration/command/RedirectMergerCommand.java
@@ -9,10 +9,14 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.concurrent.Callable;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
+
+import io.quarkiverse.roq.frontmatter.deployment.util.RoqFrontMatterTemplateUtils;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
 import picocli.CommandLine;
 import picocli.CommandLine.Command;
 import picocli.CommandLine.Parameters;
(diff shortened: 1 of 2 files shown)


migration/src/main/java/io/quarkus/tools/migration/command/RedirectMergerCommand.java
```diff
@@ -0,0 +1,247 @@
...
+                }
+            }
+        }
+
+        System.out.println("Merged " + mergedCount + " redirect(s) into target files");
+        return 0;
+    }
+
+    private boolean mergeRedirect(Path contentDir, Path redirectsDir, Path redirectFile) throws IOException {
+        String content = Files.readString(redirectFile);
```
jtama, line 79: I feel you could use [RoqFrontMatterTemplateUtils](https://github.com/quarkiverse/quarkus-roq/blob/main/roq-frontmatter/deployment/src/main/java/io/quarkiverse/roq/frontmatter/deployment/util/RoqFrontMatterTemplateUtils.java) here.

holly-cummins: You're totally right, good spot. I've done that update, and kept it as a separate commit, so you can see all the red of deleted lines. :)

jtama (approved):
Ok LGTM !

migration/src/main/java/io/quarkus/tools/migration/command/RedirectMergerCommand.java
jtama: And also thanks for removing those noisy comments.

PR #1128, opened 2026-07-23
Add capability to inject PO-translated content into Asciidoc ASTs

holly-cummins:
This PR adds capability which is the equivalent of jekyll-l10n (https://gitlab.com/CentOS/artwork/centos-web/jekyll-l10n/-/tree/v1.1.4?ref_type=tags). jekyll-l10n does one thing: AsciiDoc PO extraction and injection via an
  Asciidoctor TreeProcessor. It has two modes (update_po and translate), and runs as a Jekyll plugin during the build. I added asciidoc into the name, since it is specifically about asciidoc, and definitely not a generic implementation of #650. 

The config keys were chosen to be compatible with jekyll-l10n, and more specifically, with tsuji (https://github.com/doc-l10n-kit/tsuji), which uses them. Tsuji is used to create sites like https://pt.quarkus.io/, so support is important for us. 

If we don't have this extension, translation is still possible, but asciidoc content has to be done in one huge chunk instead of split up and managed via the adoc AST. 

The module's value is: if you have a Roq site with AsciiDoc content and PO  translation files, this plugin applies those translations at build time at the  AST level . Doing it that way has the benefit of preserving section IDs, skipping code blocks, handling tables/lists/description lists. That's useful independent of tsuji.

I didn't know what PO is, but PO is a standard format for translations. 

 If broader Roq-level l10n features are needed later (template-level, frontmatter), they'd go in a separate Quarkus extension with CDI, not in the AsciidoctorJ SPI mechanism.


holly-cummins:
This PR incorporates https://github.com/holly-cummins/quarkus-roq/pull/2 from @ynojima who very kindly tested the end to end flow with both this PR and the tsuji changes: https://quarkusio.zulipchat.com/#narrow/channel/405899-translation.2Fl10n/topic/Converting.20quarkus.2Eio.20from.20Jekyll.20to.20Roq/near/612541153

ynojima:
I tested this PR while adapting the ja.quarkus.io translation pipeline to Roq locally. A preview of the translated site is deployed at https://roq-ja-quarkusio-preview.surge.sh. I have some suggestions and will submit a PR with the changes.

  **1. Config prefix doesn't follow Quarkus/Roq conventions**

  The prefix is `l10n`, but other Roq plugins use `quarkus.roq.*` (e.g., `quarkus.roq.tagging`, `quarkus.roq.faker`). Should this be `quarkus.roq.l10n-adoc` instead? I assume the `L10N_PO_BASE_DIR` environment variable name was chosen to match tsuji's existing variable, but I'll update tsuji to match roq-plugin/l10n-adoc. Also, `L10nAdocExtensionRegistry` manually reads `System.getenv("L10N_PO_BASE_DIR")` as a fallback, but SmallRye Config automatically maps environment variables to config properties, so this shouldn't be necessary.

  **2. Logger uses `java.util.logging`**

  The rest of the Roq codebase uses `org.jboss.logging.Logger`.

  **3. Translation doesn't work for included files**

  The Treeprocessor only reads the PO file for the parent document, not for files brought in via `include::`. quarkus.io guides use `include::_includes/prerequisites.adoc[]` and similar includes for shared content, and these are left untranslated. For example, at https://roq-ja-quarkusio-preview.surge.sh/guides/telemetry-micrometer-tutorial/, the content inside the Prerequisites section remains in English. AsciidoctorJ records the included file path in `getSourceLocation().getFile()` on section nodes, so switching PO files at section boundaries addresses this.

  **4. No PO extraction (extract) capability**

  jekyll-l10n had an `update_po` mode that walked the AST to generate and update PO files. The current roq-plugin/l10n-adoc only applies  translations and has no PO generation capability. I've added an `extract-on-build` setting so that during translation, any text not found  in the PO is automatically added as a new entry. Existing translations are preserved, and entries no longer present in the AST are  removed.

  I'll submit a PR with these fixes.

ynojima:
I've submitted fixes for the points above as PRs to your branch:

- [Config prefix and env var fallback](https://github.com/holly-cummins/quarkus-roq/pull/3)
- [Logger](https://github.com/holly-cummins/quarkus-roq/pull/4)
- [Include file translation and PO extraction](https://github.com/holly-cummins/quarkus-roq/pull/5) (depends on the logger PR)

The first two are independent of each other.

roq-plugin/asciidoc-jruby/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/runtime/AsciidoctorJConverter.java
```diff
@@ -66,9 +67,11 @@ public Options createOptions(Map<String, String> asciidocAttributes, RoqTemplate
 
         final OptionsBuilder optionsBuilder = Options.builder();
         if (templateAttributes.sourcePath() != null) {
-            Path templateDir = Paths.get(templateAttributes.sourcePath()).getParent();
+            Path sourcePath = Paths.get(templateAttributes.sourcePath());
```
jtama, line 70: Isn't that also expected with asciidoctor-java ? Or is this plugin only usable with asciidoctorj, in which case I think it should be renamed.

roq-plugin/l10n-adoc/runtime/src/main/java/io/quarkiverse/roq/plugin/l10n/asciidoc/L10nAdocExtensionRegistry.java
```diff
@@ -0,0 +1,22 @@
...
+    @Override
+    public void register(Asciidoctor asciidoctor) {
+        Path poBaseDir = resolvePoBaseDir();
+        boolean extractOnBuild = L10nAdocRecorder.isExtractOnBuild();
+        asciidoctor.javaExtensionRegistry()
+                .preprocessor(new L10nAdocPreprocessor(poBaseDir))
+                .treeprocessor(new L10nAdocTreeprocessor(poBaseDir, extractOnBuild));
+    }
+
+    static Path resolvePoBaseDir() {
```
jtama: Why is this piece of code in a default package static method ?

roq-plugin/l10n-adoc/runtime/src/main/java/io/quarkiverse/roq/plugin/l10n/asciidoc/L10nAdocPoFile.java
```diff
@@ -0,0 +1,133 @@
...
+    L10nAdocPoFile() {
+        this.existingMessages = new LinkedHashMap<>();
+        this.translations = new HashMap<>();
+    }
+
+    String translate(String msgid) {
+        return translations.get(msgid);
+    }
+
+    synchronized void addEntry(String msgid) {
```
jtama: `synchronized` always feels a bit weird.

Is this needed ?

roq-plugin/l10n-adoc/runtime/src/main/java/io/quarkiverse/roq/plugin/l10n/asciidoc/L10nAdocExtensionRegistry.java
holly-cummins: You're right, it's just a wrapper, and it's not adding much. I've removed it.

roq-plugin/asciidoc-jruby/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/runtime/AsciidoctorJConverter.java
holly-cummins: This is pretty coupled to asciidoctor-java, because it hooks into the extension mechanism. I've renamed it to sit next to the adoc-jruby project.

roq-plugin/l10n-adoc/runtime/src/main/java/io/quarkiverse/roq/plugin/l10n/asciidoc/L10nAdocPoFile.java
holly-cummins: Sort of. The class we're accessing, `L10nAdocPoFile`, does need to be accessed in a threadsafe way, but expecting callers to do that isn't great. So I've replaced the `synchronized` with concurrent hashmaps in `L10nAdocPoFile` itself. I did keep `writeTo()` synchronized, because it's needed for an atomic multi-step operation.

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFile.java
```diff
@@ -0,0 +1,158 @@
...
+import org.fedorahosted.tennera.jgettext.catalog.parse.MessageStreamParser;
+
+/**
+ * Thread-safe PO file representation.
+ * Multiple tree processors may access the same cached instance concurrently.
+ */
+class L10nAdocPoFile {
+
+    private final Map<String, String> translations;
+    private final ConcurrentHashMap<String, Message> existingMessages;
```
jtama:
```suggestion
    private final Map<String, Message> existingMessages;
```

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFile.java
```diff
@@ -0,0 +1,158 @@
...
+    private final CopyOnWriteArrayList<String> encounteredMsgids;
```
jtama, line 33:
```suggestion
    private final List<String> encounteredMsgids;
```

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFile.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Adds an entry to be tracked in this PO file.
+     * Thread-safe: uses concurrent collections for lock-free operation.
+     */
+    void addEntry(String msgid) {
+        if (msgid == null || msgid.isBlank()) {
+            return;
+        }
+        // CopyOnWriteArrayList.addIfAbsent is thread-safe and atomic
```
jtama: I don't think those comments add value.

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFile.java
```diff
@@ -0,0 +1,158 @@
...
+        try (BufferedWriter bw = Files.newBufferedWriter(path)) {
+            new PoWriter().write(catalog, bw);
+        }
+        // Reset change tracking — the file now matches the in-memory state
+        existingMessages.keySet().retainAll(encounteredMsgids);
+        existingMessages.putIfAbsent("", new Message());
+        dirty.set(false);
+    }
+
+    private static Map<String, Message> parseMessages(File file) throws IOException {
```
jtama:
```suggestion
    privat Map<String, Message> parseMessages(File file) throws IOException {
```

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFile.java
```diff
@@ -0,0 +1,158 @@
...
+        while (parser.hasNext()) {
+            Message msg = parser.next();
+            if (msg.getMsgid() != null) {
+                result.put(msg.getMsgid(), msg);
+            }
+        }
+        return result;
+    }
+
+    private static Map<String, String> extractTranslations(Map<String, Message> messages) {
```
jtama:
```suggestion
    privat Map<String, String> extractTranslations(Map<String, Message> messages) {
```

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFile.java
```diff
@@ -0,0 +1,158 @@
...
+        // Use ConcurrentHashMap since translations may be read concurrently
+        Map<String, String> result = new ConcurrentHashMap<>();
+        for (Message msg : messages.values()) {
+            String msgid = msg.getMsgid();
+            String msgstr = msg.getMsgstr();
+            if (msgid == null || msgid.isEmpty()) {
```
jtama: `isEmpty` or `isBlank` as is used on `addEntry` ?

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocTreeprocessor.java
```diff
@@ -0,0 +1,312 @@
...
+                } catch (IOException e) {
+                    LOG.warnf(e, "Failed to write PO file: %s", path);
+                }
+            }
+        }
+    }
+
+    private L10nAdocPoFile loadOrCreatePoFile(Path path) {
+        return poFileCache.computeIfAbsent(path, p -> {
+            if (java.nio.file.Files.exists(p)) {
```
jtama: Why use an FQDN here ?

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocTreeprocessor.java
```diff
@@ -0,0 +1,312 @@
...
+        Path relativeSource = rootPath.relativize(sourceFile);
+        return poBaseDir.resolve(relativeSource + ".po");
+    }
+
+    private L10nAdocPoFile resolvePoFileFromSourceLocation(StructuralNode node, String rootDir) {
+        Path poPath = resolvePoPathFromSourceLocation(node, rootDir);
+        if (poPath == null) {
+            return null;
+        }
+        if (!java.nio.file.Files.exists(poPath) && !extractOnBuild) {
```
jtama: Same FQDN thingy here.

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocTreeprocessor.java
```diff
@@ -0,0 +1,312 @@
...
+        } else if (node instanceof Table table) {
+            translateTable(table, poFile);
+        } else if (node instanceof DescriptionList dlist) {
+            translateDescriptionList(dlist, poFile, rootDir, touchedPoPaths);
+        } else if (node instanceof org.asciidoctor.ast.List list) {
+            translateList(list, poFile, rootDir, touchedPoPaths);
+        } else if (node instanceof Block block) {
+            translateBlock(block, poFile, rootDir, touchedPoPaths);
+        }
+    }
```
jtama, line 201:
```suggestion
    private void processNode(StructuralNode node, L10nAdocPoFile poFile, String rootDir,
                             List<Path> touchedPoPaths) {
        return switch (node) {
            case Section section -> translateSection(section, poFile, rootDir, touchedPoPaths);
            case Table table -> translateTable(table, poFile);
            case DescriptionList dlist -> translateDescriptionList(dlist, poFile, rootDir, touchedPoPaths);
            case org.asciidoctor.ast.List list -> translateList(list, poFile, rootDir, touchedPoPaths);
            case Block block -> translateBlock(block, poFile, rootDir, touchedPoPaths);
        };
    }
```

jtama (changes requested):
Few comments here and there.

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocTreeprocessor.java
holly-cummins: No reason, good spot!

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFile.java
holly-cummins: Oh wait, this one needs the more specific type because of this line 

```
        encounteredMsgids.addIfAbsent(msgid);

```

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFile.java
holly-cummins: Agree!

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFile.java
holly-cummins: I didn't even know about that method. TIL!

holly-cummins:
Thanks for the excellent review, @jtama. I think I've fixed everything. I haven't squashed, so you can see the diff.

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFile.java
```diff
@@ -0,0 +1,156 @@
...
+                catalog.addMessage(msg);
+            }
+        }
+
+        try (BufferedWriter bw = Files.newBufferedWriter(path)) {
+            new PoWriter().write(catalog, bw);
+        }
+        // Reset change tracking — the file now matches the in-memory state
+        existingMessages.keySet().retainAll(encounteredMsgids);
+        existingMessages.putIfAbsent("", new Message());
```
jtama:
```suggestion
        existingMessages.putIfAbsent("", header);
```

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFileResolver.java
```diff
@@ -0,0 +1,24 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+
+class L10nAdocPoFileResolver {
+
+    static Optional<Path> resolve(Path poBaseDir, String baseDir, String rootDir, String docName) {
+        Path basePath = Paths.get(baseDir);
+        Path rootPath = Paths.get(rootDir);
+        Path sourcePath = basePath.resolve(docName + ".adoc");
```
jtama: in `RoqPluginAsciidocCommonProcessor`the acceptable extension for asciidoc file are `adoc` and `asciidoc`, which are indeed both often used.

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocTreeprocessor.java
```diff
@@ -0,0 +1,310 @@
...
+                    return null;
+                }
+            } else if (extractOnBuild) {
+                return new L10nAdocPoFile();
+            }
+            return null;
+        });
+    }
+
+    private Path resolvePoPathFromSourceLocation(StructuralNode node, String rootDir) {
```
jtama, line 143: isn't this like a kind of duplicated logic with the resolver ?

jtama (changes requested):
Last comments before I leave on PTO ;)

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocPoFileResolver.java
holly-cummins: Ah, good point. I'd started to refactor that constant out to so that this runtime module could access it, then I realised we actually already know the file name in some cases. So that's even easier.

roq-plugin/asciidoc-jruby-l10n/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/l10n/L10nAdocTreeprocessor.java
holly-cummins: Yes, it is. Will fix!

ia3andy (approved):
Cool! Two things I see:

---

**1. `default` branch in node type switch should not throw (`L10nAdocTreeprocessor.java`)**

```java
default -> throw new IllegalStateException("Unexpected value: " + node);
```

If AsciidoctorJ introduces a new node type or a custom extension adds one, this crashes the build. A warning + skip would be more resilient:

```java
default -> LOG.debugf("Skipping unsupported node type: %s", node.getClass().getSimpleName());
```

---

**2. PO file resolution should support classpath resources**

The PO resolver only uses filesystem paths (`Files.exists()`, `Path.toFile()`), but in Roq, content files can live in `src/main/resources/` and are resolved from the classpath. If PO files are placed alongside content in resources, the current resolver won't find them.

For reference, `AsciidocJInclude` handles this by trying classpath first via `Thread.currentThread().getContextClassLoader().getResourceAsStream()`, then falling back to filesystem. The PO resolver should follow the same pattern (at least for the translate-only path where `extractOnBuild=false`).

jtama (approved):
Nice!

holly-cummins:
Thanks @jtama and @ia3andy! I've added a commit with @ia3andy's last review comments.

holly-cummins:
(I'm confident the native CI failure in a different module is unrelated to these changes.)

PR #1130, opened 2026-07-23
fix: don't HTML-escape RSS titles/descriptions inside CDATA

philippart-s:
## Problem                                                                                                                                                                                                                       In the generated `rss.xml`, titles and descriptions containing an apostrophe or                                                                                                                                   ampersand leak to feed readers as literal HTML entities, e.g.                                                                                                                                                                    
  `Ourson&#39;s blog` or `☕ Java à l&#39;ère de l&#39;IA`. 

## Cause                                                                                                                                                                                                                         
  `fm/rss.html` wraps values in `<![CDATA[ ... ]]>`, but because the template file                                                                                                                                                 
  has a `.html` extension, Qute applies HTML escaping to the expressions. Inside                                                                                                                                                   
  CDATA the entities are never decoded, so the escaping both is unnecessary and                                                                                                                                                    
  leaks verbatim to readers.

## Fix                                                                                                                                                                                                                           
  Output the CDATA-wrapped values with `.raw` so they remain unescaped. Nullable                                                                                                                                                   
  description fields are guarded with `{#if}` because `.raw` throws on `null`                                                                                                                                                      
  whereas the previous plain expression rendered an empty string.                                                                                                                                                                  
                                                                                                                                                                                                                                   
  This also fixes `content:encoded` in full mode, whose HTML markup was being                                                                                                                                                      
  escaped inside CDATA too.                                                                                                                                                                                                        
                                                                                                                                                                                                                                   
  ## Test                                                                                                                                                                                                                          
  Adds a post with special characters (`Roq's tips & "tricks"`) to `basic-site`                                                                                                                                                    
  and asserts the RSS keeps titles/descriptions raw inside CDATA. All existing RSS                                                                                                                                                 
  tests still pass (14/14).

### DIFF
--- a/roq-frontmatter/deployment/src/test/java/io/quarkiverse/roq/frontmatter/deployment/apptest/RoqFrontMatterBasicTest.java
+++ b/roq-frontmatter/deployment/src/test/java/io/quarkiverse/roq/frontmatter/deployment/apptest/RoqFrontMatterBasicTest.java
@@ -2,6 +2,7 @@ package io.quarkiverse.roq.frontmatter.deployment.apptest;
 
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import org.junit.jupiter.api.DisplayName;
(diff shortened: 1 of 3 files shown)


jtama (approved):
LGTM

jtama:
So that made me discover that the [howto add rss](https://iamroq.dev/posts/rss-feed-of-your-blog-posts/), has an alias that points to RSS, which means rss is not availabe on iamroq.

I'll create an issue tomorrow.

Thanks for the PR.

jtama:
Yeah ok, no that was dumb. We do, the url is https://iamroq.dev/rss.xml, and not https://iamroq.dev/rss.

PR #1134, opened 2026-07-24
Expand docs for dynamic selection and add supporting tests

holly-cummins:
The use case for dynamic selection, or how it could be used to support remote sources, wasn't clear to me when I read the docs. I'm hoping this makes the range of use cases a bit clearer. 

Because I had code examples, I also added some integration tests exercising similar code.

### DIFF
--- a/docs/modules/ROOT/pages/quarkus-roq-generator.adoc
+++ b/docs/modules/ROOT/pages/quarkus-roq-generator.adoc
@@ -41,47 +41,260 @@ Check how the Roq blog is deployed on GitHub pages https://github.com/quarkivers
 
 == Advanced Usage
 
-=== Configure the static selection
+=== Understanding selections and sources
 
-You can configure application paths which should be generated as static files from the Quarkus configuration:
+Roq Generator works by making HTTP requests to your running Quarkus application and saving the responses as static files.
+You control **which pages to generate** through *selections*.
+
+For typical Roq sites using file-based content (Markdown/AsciiDoc in `content/`), no selection configuration is needed - the
+Roq extension automatically discovers and generates your pages.
+
+**Where content comes from:**
+
+* **File-based content** - Markdown/AsciiDoc files in your `content/` directory (the default Roq pattern)
+* **Database content** - Blog posts, products, or other content stored in PostgreSQL, MySQL, etc.
+* **Remote APIs** - Content fetched from headless CMS (Contentful, Strapi), external APIs, or microservices
+* **Other** - If you can code it, you can create it
+
+**How you specify what to generate (selections):**
+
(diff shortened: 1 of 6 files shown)


docs/modules/ROOT/pages/quarkus-roq-generator.adoc
```diff
@@ -41,47 +41,260 @@ Check how the Roq blog is deployed on GitHub pages https://github.com/quarkivers
...
+**How you specify what to generate (selections):**
+
+* **Static selection** - Fixed paths configured in `application.properties` (simple, declarative)
+* **Dynamic selection** - Paths computed at build time from databases or APIs (programmatic, flexible)
+
+The choice between static and dynamic selection depends on whether you know all your URLs at configuration time or need to discover them at build time.
+
+=== Static selection (configuration-based)
+
+Use static selection when you have want to include a **fixed set of extra paths** that don't change based on external data.
```
jtama: `when you have want to include`, I know it's a difficult one, but you have to choose between _have_ and _want_.

docs/modules/ROOT/pages/quarkus-roq-generator.adoc
```diff
@@ -41,47 +41,260 @@ Check how the Roq blog is deployed on GitHub pages https://github.com/quarkivers
...
+    ContentApiClient apiClient; // <1>
+
+    @Produces
+    @Singleton
+    RoqSelection produce() {
+        // Fetch content list from remote API at build time
+        List<Post> posts = apiClient.getAllPosts(); // <2>
+
+        // Generate a static page for each post
+        List<SelectedPath> paths = posts.stream()
```
jtama: Here you collect the list of posts to generate later, but I still don't get when the content is actualy retrieved

docs/modules/ROOT/pages/quarkus-roq-generator.adoc
holly-cummins: Fine! Insist that the docs be grammatically correct! See if I care!

docs/modules/ROOT/pages/quarkus-roq-generator.adoc
holly-cummins: Yeees, I see what you mean. It gets really confusing between "a url that is actually remote on a remote site" and "all the pages we bake into the final site start life as urls that we download," and for testability, the example used a local-remote page, but the local-remote path is also just the normal-roq path. I'll rewrite to make it less testable but less confusing and less not-actually-useful.

holly-cummins: It's all a bit confusing. But I *think* I've made the example clearer.

jtama (approved):

PR #1150, opened 2026-08-10
Refactor RoqUrl methods to support multiple paths in isActive and remove nav method

matheusandre1:
Follow up review : https://github.com/quarkiverse/quarkus-roq/pull/1041#issuecomment-5067290437


ia3andy (commented):
Add null guards on `path` and varargs elements, `StringPaths.removeTrailingSlash(null)` will NPE if a Qute expression resolves to null:

```java
if (path != null && current.equals(StringPaths.removeTrailingSlash(path))) {
```

and:

```java
if (p != null && current.equals(StringPaths.removeTrailingSlash(p))) {
```

ia3andy (approved):

PR #1158, opened 2026-08-20
Asciidoc: Use absolute paths when setting xrefpathprefix

holly-cummins:
I noticed that xrefs work on the generated site, but are sometimes failing in dev mode. The problem turned out to be because in the generated site, links always end up with a trailing slash (`/guides/my-page/`), but in dev mode, the trailing slash is not there. This means that the url gets resolved to `/guides/my-page../something`, which then turns into `/something`, which doesn't exist. Instead of hardcoding `..`, we can work out the actual parent path and use that as the relative prefix. 

I always get a bit muddled about trailing slashes and web generation so I wasn't totally sure this fix is better than just adding the trailing slash, but two different claude models and gemini all agreed that it was cleaner, more semantically clear, and more idiomatic than using `..` in urls. So I've gone with that. 

I added some utility methods along the way, so I converted a few other classes to use them as well.

### DIFF
--- a/roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/model/Page.java
+++ b/roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/model/Page.java
@@ -3,6 +3,7 @@ package io.quarkiverse.roq.frontmatter.runtime.model;
 import static io.quarkiverse.roq.frontmatter.runtime.RoqFrontMatterKeys.DESCRIPTION;
 import static io.quarkiverse.roq.frontmatter.runtime.RoqFrontMatterKeys.TITLE;
 import static io.quarkiverse.roq.frontmatter.runtime.RoqTemplates.ROQ_PAGE_CONTENT_FRAGMENT;
+import static io.quarkiverse.roq.frontmatter.runtime.model.RoqUrl.isAbsolute;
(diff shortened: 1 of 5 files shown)


holly-cummins:
By the way, I have a workaround for this so it's not a blocker for the quarkusio migration.

ia3andy (commented):
Nice fix, the core approach (absolute `relfileprefix` from page URL) is solid. A few suggestions:

1. **Rename `collectionBasePath` variable**: this also applies to standalone pages (e.g. `/about` or `/docs/faq`), not just collection documents. Something like `basePath` would be more accurate.

2. **Add unit tests for `parentPath`**: this is a utility method with edge cases (trailing slashes, single-segment paths, root). It deserves its own test coverage rather than only being tested indirectly through the integration test.

3. **Duplicate test**: `testXrefUsesAbsolutePaths` asserts the same `href="/guides/rabbitmq/"` string that `testMyDoc` already checks. The new method can be dropped since the existing assertion covers it.

holly-cummins:
Suggestions all make sense, all implemented now in a follow-up commit.

holly-cummins:
It would be great to get this into the next release, it's causing a few broken links on the live site. I worked round a bunch with asciidoc config, but there's a few that I can't work round.

ia3andy (approved):
LGTM

PR #1159, opened 2026-08-20
Support both HEAD and GET in dev mode

holly-cummins:
I noticed that some of my link crawling tests are failing in dev mode, because they use HEAD for efficiency. The current dev mode router only handles `GET`. I've added the `HEAD` handling, and also some tests.


holly-cummins:
I can work around this, so this issue shouldn't block anything.

ia3andy (approved):

PR #1164, opened 2026-08-21
Correct generated files

holly-cummins:
Release CIs fail if the build alters any files. 

https://github.com/quarkiverse/quarkus-roq/pull/1162 did a fix to commit some generated files that were blocking the release, but I must have been working on a state checkout, and so I created duplicated entries. I'm not sure why CI is generating files differently than local builds, because I still can't reproduce the source changes locally, but that's a broader problem for another day. :)


holly-cummins:
This was needed because I'd run builds without doing `clean`. The config class was moved from deployment to runtime in commit 856671ab. On my machine, the **old** deployment target still had stale annotation processor output from when RoqTaggingConfig.java lived in deployment/. In combination with the new output, there was a duplicate entry.

PR #1174, opened 2026-08-25
fix(lunr): make search dialog and trigger accessible

and-y:
## Summary

The Lunr search overlay and trigger button had no accessibility semantics:

- The overlay had no `role="dialog"`, `aria-modal`, or accessible name, so screen readers didn't announce it as a dialog.
- The close button (`×`) and the search input had no accessible name (the input relied on `placeholder` only).
- The search results region had no `aria-live`, so results appearing as you type were never announced.
- The default `search-button` and `search-trigger-input` components were plain `<div>`s with no keyboard support (not focusable, not activatable with Enter/Space).
- Closing the overlay dropped focus instead of returning it to the element that opened it, and Tab could escape the dialog into the page behind it.

## Changes

- `search-overlay.html`: `role="dialog"`, `aria-modal="true"`, `aria-label="Search"`, `aria-hidden` on the overlay; `aria-label` on the close button and input; `role="region"` + `aria-live="polite"` on the results container.
- `search-button.html` / `search-button-input.html`: changed from `<div>` to `<button type="button">` with `aria-label`, plus matching CSS resets in `search.css` so native button chrome doesn't regress the existing look.
- `search.js`: toggles `aria-hidden` on open/close, saves `document.activeElement` before opening and restores focus to it on close, and traps `Tab` within the dialog while it's open.

## Test plan

- Added `RoqPluginLunrTagsAccessibilityTest` (deployment module) with a minimal site fixture, asserting the rendered markup carries the new ARIA attributes.
- Ran `mvn verify` for `roq-plugin/lunr/runtime` and `roq-plugin/lunr/deployment` — all tests pass (17 existing + 2 new).
- Manually verified the fix on a real site consuming this plugin: overlay announces as a dialog, results are announced live, and the trigger button is keyboard-operable.

### DIFF
--- a/roq-plugin/lunr/deployment/src/test/java/io/quarkiverse/roq/plugin/lunr/test/RoqPluginLunrTagsAccessibilityTest.java
+++ b/roq-plugin/lunr/deployment/src/test/java/io/quarkiverse/roq/plugin/lunr/test/RoqPluginLunrTagsAccessibilityTest.java
@@ -1,0 +1,45 @@
+package io.quarkiverse.roq.plugin.lunr.test;
+
+import static org.hamcrest.Matchers.containsString;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusExtensionTest;
+import io.restassured.RestAssured;
+
+/**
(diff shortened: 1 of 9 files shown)


ia3andy (approved):
Thanks for this nice contribution!

ia3andy:
@all-contributors please add @and-y for code

PR #1177, opened 2026-08-26
fix(frontmatter): omit llms.txt/llms-full.txt heading for collections with no visible docs

and-y:
## Summary

A collection whose every document opts out via `llmstxt: false` still printed
an empty `## <Collection>` heading in `/llms.txt` and `/llms-full.txt`. The
heading was gated only on `!collection.hidden && !collection.derived`, never
on whether any document in the collection actually passed the per-doc
`llmstxt` filter used inside the loop. A fully-opted-out collection therefore
rendered a dangling heading with nothing under it.

Example, with a `recipes` collection whose layout sets `llmstxt: false`:

```
## Recipes

## Pages

- [About](/about/): About this site
```

## Fix

- `RoqLlmsTxtTemplateExtension`: add `hasLlmstxtEntries(RoqCollection)`,
  returning whether at least one document in the collection isn't opted out.
- `templates/fm/llms.html` / `llms-full.html`: require
  `collection.hasLlmstxtEntries` alongside the existing
  `!collection.hidden && !collection.derived` condition.

## Test plan

- New `RoqFrontMatterLlmsTxtEmptyCollectionTest` (fixture:
  `llms-empty-collection-site`) covers both directions:
  - a collection whose only doc opts out via `llmstxt: false` → heading omitted
  - a sibling collection with a visible doc → heading and entry still printed
- Verified against unmodified code first to confirm the reproduction, then
  against the fix.

### DIFF
--- a/roq-frontmatter/deployment/src/test/collection-site/content/guides/getting-started.html
+++ b/roq-frontmatter/deployment/src/test/collection-site/content/guides/getting-started.html
@@ -2,6 +2,7 @@
 layout: guide
 title: Getting Started (Local Override)
 slug: getting-started
+llmstxt: false
 ---
 <h1>{page.title}</h1>
 <p>This is the local version of the getting started guide, which should override the resource version.</p>
--- a/roq-frontmatter/deployment/src/test/collection-site/content/llms-full.qute.txt
+++ b/roq-frontmatter/deployment/src/test/collection-site/content/llms-full.qute.txt
@@ -1,0 +1,1 @@
+{#include fm/llms-full.html}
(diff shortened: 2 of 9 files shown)


ia3andy (commented):
Nice fix! Two suggestions:

1. **Javadoc on `hasLlmstxtEntries`**: the double negative ("not opted out") is a bit confusing. Something like "Returns whether the collection has at least one document visible in llms.txt" would read better.

2. **Test fixture**: a whole new site seems overkill for this. Could you add a `llmstxt: false` document to an existing test fixture and assert the heading is omitted there? That would avoid the extra test class + 6 new files.

and-y:
Yes, i will have a look into it.

and-y:
Please have a look again. Addressed both points:

- Reworded hasLlmstxtEntries javadoc.
- Dropped the new test class/fixture and reused collection-site.

ia3andy (approved):
LGTM

ia3andy:
@all-contributors please add @and-y for code

PR #1180, opened 2026-08-26
Fix adoc tag processing

### LINKED ISSUE #1178: Asciidoc processor uses wrong separator for tags
See https://github.com/quarkusio/quarkusio.github.io/issues/2939.

PR incoming.

holly-cummins:
Resolves #1178

AsciiDoc's spec uses ; as the separator, but the Ruby (Asciidoctor) implementation actually accepts both ; and ,. The version in Roq only supported commas. This isn't ideal, since the canonical syntax (and what the Quarkus guides use) is  semicolons.

The fix handles both: tagsValue.split("[;,]"). It also adds tests for both.

With the original ,-split code,` tags=example;!ignore;!registry` was treated as a single tag name `"example;!ignore;!registry"`, which matched nothing, so the result was 0 lines.

### DIFF
--- a/roq-plugin/asciidoc-jruby/runtime/pom.xml
+++ b/roq-plugin/asciidoc-jruby/runtime/pom.xml
@@ -31,6 +31,17 @@
             <artifactId>asciidoctorj-diagram</artifactId>
             <version>${asciidoctorj-diagram.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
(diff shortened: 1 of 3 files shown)


ia3andy (approved):
Nice work! The rewrite from sequential single-tag extraction to single-pass with proper include/exclude semantics is a solid improvement.

Minor: `resolveTagsValue` is called twice (once in `pushInclude` for the warning, once inside `processInclude`). Could pass the resolved value through instead of re-resolving from attributes.

ia3andy:
I am merging, you can always do a followup PR for the comments

PR #1182, opened 2026-08-26
Use 'true' for boolean asciidoc config

holly-cummins:
This is related to https://github.com/quarkiverse/quarkus-roq/pull/1175, but I think the solution here is more natural. For an asciidoc config value that's either present or not, we can't just put it loose into the Roq config. But we can say 'true' in the config, and that seems semantically natural.  

For example, 

#### Jekyll _config.yml:
```
  asciidoctor:
    attributes:
      source-highlighter: highlightjs
      sectanchors: ''
      icons: font
```
becomes

 #### Roq application.properties:
 ```
  quarkus.asciidoc.attributes.source-highlighter=highlightjs
  quarkus.asciidoc.attributes.sectanchors=true
  quarkus.asciidoc.attributes.icons=font
````

I notice even Jekyll was a bit awkward, because it had to do an empty string. Only it also had an option for an array syntax, where presence was enough: https://github.com/asciidoctor/jekyll-asciidoc

### DIFF
--- a/blog/content/docs/migrating.adoc
+++ b/blog/content/docs/migrating.adoc
@@ -831,6 +831,9 @@ If you are migrating manually, consider adding them to `config/application.prope
 
 | `quarkus.web-bundler.bundling.external=/assets/*`
 | Tells the CSS bundler not to resolve Jekyll-style `/assets/...` URL references
+
+| `quarkus.asciidoc.attributes.sectanchors=true`
+| Enables section anchor links (use `true` for any boolean AsciiDoc attribute since SmallRye Config does not support empty-string map values; Roq converts `true` to an empty string before passing to AsciidoctorJ)
 |===
 
 === LLM prompt for configuration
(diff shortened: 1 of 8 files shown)


ia3andy (commented):
Good fix. The `true` → `""` conversion for boolean AsciiDoc attributes is a clean workaround for the SmallRye Config limitation. The non-JRuby `asciidoc` plugin has no attribute config handling, so only the JRuby plugin needs this.

A few notes:

1. **`looksLikeValueAttribute` is fragile.** A suffix heuristic will miss edge cases. Worth adding a comment that this is best-effort for the migration tool, not a general-purpose classifier.

2. **`icons` edge case.** An empty `icons: ''` in Jekyll config would be emitted as `icons=true`, which becomes `icons=""` in AsciidoctorJ. That happens to be correct (`:icons:` enables default icons), but it's by accident since `icons` can also take values like `font`.

3. **`System.out.println` in migration tool.** Minor: the skip message uses stdout while other parts might use a logger. Not a blocker.

4. **Tests are solid.** Good coverage of the converter, the migration heuristic, and both boolean vs value attribute paths.

PR #1183, opened 2026-08-26
[Breaking] Allow placeholders to support dots in paths

holly-cummins:
At the moment, Page pages can have dots in their paths, but things using placeholder names cannot. I noticed this in the Roq migration, when /blog/vscode-quarkus-1.1.0/ stopped working. 

I've updated the placeholder to also allow dots. I wondered about making it configurable, but I think that's adding user complexity for not much gain. `PageFiles.slugifyFile` uses `allowDots=true` with the explicit rationale about version numbers. Making `:Name/:name` match is consistent with that.

Also, dots in slugs are standard. Jekyll and Hugo both preserve them. 

More generally, there's not really a use case for stripping dots, because if someone wanted dashes, they could use dashes in the file name. 

The one caution is that: if someone has already built their site on Roq with the current behavior, changing the default would move their URLs. So this would want a release note.

### DIFF
--- a/roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/utils/TemplateLink.java
+++ b/roq-frontmatter/runtime/src/main/java/io/quarkiverse/roq/frontmatter/runtime/utils/TemplateLink.java
@@ -122,7 +122,7 @@ public class TemplateLink {
             name = pageSource.baseFileName();
         }
 
-        return StringPaths.slugify(removeDate(name), true, false);
+        return StringPaths.slugify(removeDate(name), true, true);
     }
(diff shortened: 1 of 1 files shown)


ia3andy:
Good fix, but a bit delicate..

holly-cummins:
> Good fix, but a bit delicate..

Do you have suggestions for making it less delicate?

ia3andy:
not much :-/ but a least let's wait for a minor?

holly-cummins:
> not much :-/ but a least let's wait for a minor?

Sounds fair!

ia3andy:
Ok I am merging, we should add the release note

PR #1184, opened 2026-08-26
Bump yupiik to pick up header parsing fix

### LINKED ISSUE #1136: AsciiDoc document title (`= Title`) is not used as page title
In Jekyll with jekyll-asciidoc, the AsciiDoc document title (`= Title`) is automatically used as the page title, even without a `title` in the YAML front matter.

In Roq, `page.title()` only reads `title` from the front matter and falls back to the source file path:

```java
public String title() {
    return data().getString(TITLE, sourcePath());
}
```

This is a behavioral difference that causes problems when migrating Jekyll-based sites to Roq. For example, [quarkusio.github.io](https://github.com/quarkusio/quarkusio.github.io) has many AsciiDoc guide pages (e.g., [security-jwt.adoc](https://github.com/quarkusio/quarkusio.github.io/blob/ac0ccdc088d64864a6645034eb91e24d12b59b39/_guides/security-jwt.adoc)) that define their title using AsciiDoc syntax (\`= Using JWT RBAC\`) without a front matter \`title\`. With Roq, \`page.title()\` returns the source file path instead of the document title.

holly-cummins:
Resolves #1136 

As well as the version bump which picks up @jtama's fix, one API changed, so I did an update for that. We may want to implement support for multiple authors, now that it's available.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
         <asciidoctorj.version>3.0.1</asciidoctorj.version>
         <asciidoctorj-diagram.version>3.2.1</asciidoctorj-diagram.version>
         <quarkus-web-bundler.version>2.3.3</quarkus-web-bundler.version>
-        <asciidoc-java.version>1.2.14</asciidoc-java.version>
+        <asciidoc-java.version>1.2.16</asciidoc-java.version>
         <jandex.version>3.6.0</jandex.version>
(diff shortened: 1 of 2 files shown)


jtama (approved):
About multiple authors, it can already be done, but using frontmatter.
Also the default theme doesn't handle them, but it's an easy fix.

Issue #158, opened 2024-10-09
Bug when using --- inside a markdown code section


mcruzdev:
When using the following inside a blog post as markdown:

When you are using `- - - `block.

![image](https://github.com/user-attachments/assets/7428bd38-f18b-4559-a4a1-1837c51f5431)

The page looks like:

Without page content.

![image](https://github.com/user-attachments/assets/ac88e24b-c030-4a35-b1aa-db942dba2814)

ia3andy:
I think there is an issue with the frontmatter regex, I need to fix it

PR #1202, opened 2026-09-02
Correct getParentPath to not mangle urls when they have a host and scheme

holly-cummins:
#158 was the right idea, but it had an unfortunate bug that turned up in production; the `getParentPath()` method fails totally for full urls – ie, when `site.url` is set. Given a full URL it left the scheme and host in place, so the computed `relfileprefix` became /https://host/guides/ and every inter-document xref resolved to a 404.

I've checked this preserves the intended behaviour of #158 with fewer appalling side effects. I've also added some unit tests for the missing scenario, and a slightly higher level test which simulates the `site.url` scenario.


ia3andy (approved):
Thanks

PR #100, opened 2024-09-20
Fix #99: Link defensive null check

melloware:
Fix #99: Link defensive null check

This gets past the NPE but now I get this..

```
ERROR] Failed to execute goal io.quarkus.platform:quarkus-maven-plugin:3.14.4:build (default) on project roq-blog: Failed to build quarkus application: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
[ERROR]         [error]: Build step io.quarkus.arc.deployment.SyntheticBeansProcessor#initStatic threw an exception: java.lang.IllegalStateException: A synthetic bean with identifier I8e6VDWFfp-mJuk3LsVFwebeYF0 is already registered: SYNTHETIC bean [types=[io.quarkiverse.roq.frontmatter.runtime.Page, java.lang.Object], qualifiers=[@Default, @Any, @Named("/")], target=n/a]
[ERROR]         at io.quarkus.arc.processor.BeanDeployment.addSyntheticBean(BeanDeployment.java:1511)
[ERROR]         at io.quarkus.arc.processor.BeanDeployment$BeanRegistrationContextImpl.accept(BeanDeployment.java:1866)
[ERROR]         at io.quarkus.arc.processor.BeanDeployment$BeanRegistrationContextImpl.accept(BeanDeployment.java:1835)
[ERROR]         at io.quarkus.arc.processor.BeanConfigurator.done(BeanConfigurator.java:114)
[ERROR]         at io.quarkus.arc.deployment.SyntheticBeansProcessor.configureSyntheticBean(SyntheticBeansProcessor.java:92)
[ERROR]         at io.quarkus.arc.deployment.SyntheticBeansProcessor.initStatic(SyntheticBeansProcessor.java:39)
[ERROR]         at java.base/java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:733)
[ERROR]         at io.quarkus.deployment.ExtensionLoader$3.execute(ExtensionLoader.java:856)
[ERROR]         at io.quarkus.builder.BuildContext.run(BuildContext.java:256)
[ERROR]         at org.jboss.threads.ContextHandler$1.runWith(ContextHandler.java:18)
[ERROR]         at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2516)
[ERROR]         at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2495)
[...]

PR #1206, opened 2026-09-03
Handle attributes in asciidoc headers

holly-cummins:
Ideally this would be done in the Yupiik layer, so I've also opened https://github.com/yupiik/tools-maven-plugin/issues/100. If that one closes, we'd still want the tests from this PR, just a bit less of the code. 

See https://github.com/quarkusio/quarkus/pull/56389

### DIFF
--- a/roq-plugin/asciidoc-common/deployment/src/main/java/io/quarkiverse/roq/plugin/asciidoc/common/deployment/AsciidocHeaderParser.java
+++ b/roq-plugin/asciidoc-common/deployment/src/main/java/io/quarkiverse/roq/plugin/asciidoc/common/deployment/AsciidocHeaderParser.java
@@ -55,6 +55,20 @@ public class AsciidocHeaderParser {
     // Matches attribute entries: :name: value, :!name: (unset), :name!: (unset)
     private static final Pattern ATTRIBUTE_ENTRY = Pattern.compile("^:(!?\\w[^:]*):(?:[ \\t]+(.*))?$");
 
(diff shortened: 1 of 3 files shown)


ia3andy:
I am holding on merging this?

holly-cummins:
> I am holding on merging this?

Ummm. I think either is fine, but #100 is still open (mostly because I haven't provided a PR :) ), so I think maybe we should merge this, and then remove the implementation (but not the tests) next time we update yupiik. But either is fine.

ia3andy (commented):
**Overall:** Good addition. Resolving attribute references in title/description/image is a real gap that this fills. The tests are thorough and the TODO comments for Yupiik upstream are helpful.

**Two suggestions to consider:**

1. **Resolve in `processAttributes()` instead of at each call site.** Asciidoctor resolves attribute references at definition time inside `set_attribute` &rarr; `apply_attribute_value_subs` (which applies `HEADER_SUBS = [:specialcharacters, :attributes]`). Doing the same in `processAttributes()` (resolving each value against the attributes already stored so far) would:
   - Handle `page-*` values automatically (currently left unresolved)
   - Handle chained references (`:base: /assets` &rarr; `:imagesdir: {base}/images` &rarr; `:image: {imagesdir}/hero.png`)
   - Remove the three `resolveAttributeReferences` call sites in `toPageData()`

2. **Consolidate the duplicate TODOs** (lines 58-63 and 133-140) about `yupiik#100` into one.

PR #1212, opened 2026-09-04
fix: resolve <html lang> from page/site frontmatter instead of hardcoded "en"

and-y:
## Summary
- The `roq-base`, `roq-resume`, and `roq-linktree` theme layouts hardcoded `<html lang="en">`.
- Roq already resolves a page/site `lang` frontmatter key for the `og:locale` SEO meta tag (`seo.html` / `seoLocale.html`), so this reuses the same `page lang -> site lang -> JVM default locale` fallback chain for the `<html lang="...">` attribute.
- Documents the `lang` frontmatter key in `RoqFrontMatterKeys` and the `roq-frontmatter` skill doc.

## Test plan
- [x] New `RoqFrontMatterBaseThemeTest` assertions cover page-lang and page->site fallback via the `base-theme-site` fixture.
- [x] New `RoqFrontMatterHtmlLangGlobalFallbackTest` covers the JVM-default fallback (no `lang` set anywhere).
- [x] Added `<html lang>` smoke assertions to the existing `roq-resume` and `roq-linktree` theme integration tests.

### DIFF
--- a/roq-theme/linktree/runtime/src/main/resources/templates/theme-layouts/roq-linktree/default.html
+++ b/roq-theme/linktree/runtime/src/main/resources/templates/theme-layouts/roq-linktree/default.html
@@ -1,7 +1,8 @@
 {@io.quarkiverse.roq.frontmatter.runtime.model.Page page}
 {@io.quarkiverse.roq.frontmatter.runtime.model.Site site}
+{#let pageLang=page.data.lang?? siteLang=site.data.lang??}
 <!DOCTYPE html>
-<html lang="en" class="bg-slate-100 dark:bg-slate-900">
+<html lang="{pageLang ?: siteLang ?: global:locale}" class="bg-slate-100 dark:bg-slate-900">
 <head>
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
@@ -13,3 +14,4 @@
   {#insert /}
 </body>
 </html>
+{/let}
--- a/roq-theme/linktree/runtime/src/main/resources/templates/theme-layouts/roq-linktree/default.html
+++ b/roq-theme/linktree/runtime/src/main/resources/templates/theme-layouts/roq-linktree/default.html
@@ -1,8 +1,7 @@
 {@io.quarkiverse.roq.frontmatter.runtime.model.Page page}
 {@io.quarkiverse.roq.frontmatter.runtime.model.Site site}
-{#let pageLang=page.data.lang?? siteLang=site.data.lang??}
 <!DOCTYPE html>
-<html lang="{pageLang ?: siteLang ?: global:locale}" class="bg-slate-100 dark:bg-slate-900">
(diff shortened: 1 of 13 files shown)


roq-theme/linktree/runtime/src/main/resources/templates/theme-layouts/roq-linktree/default.html
```diff
@@ -1,7 +1,8 @@
 {@io.quarkiverse.roq.frontmatter.runtime.model.Page page}
 {@io.quarkiverse.roq.frontmatter.runtime.model.Site site}
+{#let pageLang=page.data.lang?? siteLang=site.data.lang??}
 <!DOCTYPE html>
```
ia3andy, line 4: I wonder why it's not using the base theme

roq-theme/resume/runtime/src/main/resources/templates/theme-layouts/roq-resume/default.html
```diff
@@ -1,9 +1,13 @@
+{@io.quarkiverse.roq.frontmatter.runtime.model.Page page}
+{@io.quarkiverse.roq.frontmatter.runtime.model.Site site}
+{#let pageLang=page.data.lang?? siteLang=site.data.lang??}
 <!DOCTYPE html>
-<html lang="en">
```
ia3andy, line 2: same here

roq-theme/linktree/runtime/src/main/resources/templates/theme-layouts/roq-linktree/default.html
and-y: You're suggesting `roq-resume` and `roq-linktree` should extend `roq-base/default` (like `roq-default` does via `layout: roq-base/default`) instead of each defining their own `<html>` boilerplate. Is that right?

If so, should I fold that refactor into this PR (it'd let me drop the `lang` changes from these two files, since they'd inherit it from `roq-base`), or should I open a follow-up issue for the `roq-base` extraction and keep this PR scoped to the `lang` fix?

ia3andy (approved):

ia3andy (changes requested):
Ah, I've found an issue:
- global:locale is Locale.getDefault(). Qute renders it via .toString(), which produces underscore-separated tags like en_US. The HTML spec requires BCP 47 language tags, which use hyphens (en-US).
-  {#let} wrapping the entire document instead of just the right part

For the base theme stuff, that can be another PR. 

Sorry for the delay!

and-y:
Fixed both:

- Added a separate `global:htmlLocale` (`Locale#toLanguageTag()`, e.g. `en-US`) for the `<html lang>` fallback, keeping `global:locale` (`Locale#toString()`, e.g. `en_US`) as-is for `og:locale`.
- Dropped the `{#let}` wrapper: `JsonObject#getString("lang")` returns `null` for a missing key instead of throwing, so `page.data.getString('lang') ?: site.data.getString('lang') ?: global:htmlLocale` works inline in the attribute without a scoped binding.

Opened #1222 for the `roq-base` extraction.

ia3andy (approved):
Thanks that's way better!

Issue #1223, opened 2026-09-09
Advertise each page's Markdown twin with a link rel="alternate" type="text/markdown", only when the twin was generated


rolfedh:
Follow-up to #1105, agreed [there](https://github.com/quarkiverse/quarkus-roq/pull/1105#issuecomment-5596799702): the pull request publishes the twins, this issue makes them discoverable from the page.

Each HTML page whose Markdown twin exists gets

```html
<link rel="alternate" type="text/markdown" href="<page-url>.md">
```

in its head, so agents and tools that look for a Markdown alternative find it without guessing the URL.

Two requirements:

- The tag is driven by the set of twins actually generated, never by the page type. A page whose conversion failed (the plugin logs a warning and writes no file) must not advertise a twin.
- A test proves it: a page whose twin fails carries no alternate link, and a page with a twin does.

Note for the implementation: the twin step runs after page data is frozen (data modifications are applied in step 2, page templates exist from step 3), so the twin build step cannot stamp the twin URL into page data. The likely shape is a registry of the pages that got a twin, recorded at build time and exposed to templates as something like `page.markdownTwin` (the URL, or null), plus the head snippet that reads it. The llms.txt pointer (#1219) can read the same registry.

Issue #1219, opened 2026-09-08
Point llms.txt links at the Markdown twins (follow-up to #1105)


rolfedh:
Follow-up to #1104 and #1105, agreed with Holly Cummins on the PR (https://github.com/quarkiverse/quarkus-roq/pull/1105#issuecomment-5586764308).

### What

Once the markdown-twin plugin is on the classpath, the links in `llms.txt` should point at the `.md` twin of each page (`<page-url>.md`) instead of the HTML page, falling back to the HTML URL for pages that have no twin (HTML pages, pages with `llmstxt: false`).

### Why

`llms.txt` is a directory for agents, and Markdown is what they fetch; the twin exists for exactly that. The little real traffic llms.txt files receive comes from coding agents that follow per-page links.

### Constraint found while looking at the mechanism

The twin build step consumes `RoqFrontMatterPageTemplateBuildItem`, which exists from Step 3, while `RoqFrontMatterDataModificationBuildItem` is applied in Step 2, so the twin step cannot stamp the twin URL into page data without a build-item cycle. The likely shape is a runtime template extension (for example `page.markdownTwin`) that derives the twin URL from `page.url` and the `llmstxt` flag, used by `templates/fm/llms.html` with the HTML URL as the fallback; it must resolve to the HTML URL when the twin plugin is absent.

### Out of scope, decided on #1105

- `llms-full.txt` stays as it is (rarely fetched; see the discussion on the PR).
- `sitemap.xml` stays as it is: the twins are alternates of existing pages, not pages.
- The `<link rel="alternate" type="text/markdown">` tag per HTML page is tracked on #1104.

PR #1224, opened 2026-09-09
Advertise each page's Markdown twin with a link rel="alternate" type="text/markdown", only when the twin was generated

rolfedh:
Draft on top of #1105 until that one merges: the first commit here is 1105's, the change of this pull request is the second commit (`aa306a37`). I rebase it and mark it ready once 1105 is in.

Implements #1223, the follow-up agreed on 1105 ([comment](https://github.com/quarkiverse/quarkus-roq/pull/1105#issuecomment-5596799702)).

**What it does**

- The `seo` tag gains a `{#seoAlternates page site /}` hook with an empty default tag, shaped like its other sub-tags.
- The markdown-twin plugin overrides that hook, the same way og-card overrides `seoImage` (a higher-priority template path), with `<link rel="alternate" type="text/markdown" href="<page-url>index.md" />`, rendered only when `page.markdownTwin` is set.
- `page.markdownTwin` is a template extension backed by a registry the processor records at static init: page resource path to twin path, for the pages that really got a file. A page that opted out (`llmstxt: false`) or whose conversion failed and got no file is absent, so no page can advertise a twin that does not exist.

**Tests** (`RoqBlogTest`): `/docs/basics/` carries the link; the opted-out page and a new fixture whose conversion fails on purpose (`markups/twin-broken.adoc`, a malformed `ifeval` the in-process converter rejects while Asciidoctor renders the page) carry neither the link nor a twin.

Not in this pull request: the llms.txt pointer (#1219), which can read the same registry.


rolfedh:
@holly-cummins @ia3andy a heads-up before you click around this preview.

On the surge preview, a docs URL typed without its trailing slash returns the Markdown twin instead of the page:

- https://quarkiverse-roq-blog-1224-preview.surge.sh/docs/basics/ is the HTML page.
- https://quarkiverse-roq-blog-1224-preview.surge.sh/docs/basics is served as `text/markdown`, byte for byte the twin at `/docs/basics.md`.

Surge resolves an extensionless path to `basics.md` before `basics/index.html`. Pages without a twin behave as before: `/docs/plugins` redirects to `/docs/plugins/`. Every link the site emits carries the slash, so this only bites hand-typed URLs. The #1105 preview has the same behaviour, since it also ships the twins. The live site on GitHub Pages currently redirects `/docs/basics` to `/docs/basics/`; I have not checked what GitHub Pages does once a `basics.md` sits beside `basics/index.html`.

PR #1225, opened 2026-09-10
Add sort virtual method to jekyll templates

holly-cummins:
Retroactively fixing a bug I discovered post-conversion. The methods in the jekyll template are missing a sort in some scenarios.

### DIFF
--- a/migration/src/main/java/io/quarkus/tools/migration/JekyllFiltersExtension.java
+++ b/migration/src/main/java/io/quarkus/tools/migration/JekyllFiltersExtension.java
@@ -23,7 +23,9 @@ import io.vertx.core.json.JsonObject;
 @TemplateExtension
 public class JekyllFiltersExtension {
 
-    // Null-safe get — prevents NPE when key is null (e.g. from ?? operator)
+    /**
+     * Null-safe get for JsonObject. Prevents NPE when key is null (e.g. from ?? operator).
+     */
     static Object get(JsonObject obj, String key) {
         if (obj == null || key == null || key.isEmpty()) {
             return null;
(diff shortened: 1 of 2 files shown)


migration/src/main/java/io/quarkus/tools/migration/JekyllFiltersExtension.java
```diff
@@ -129,6 +178,41 @@ static List<?> sort(List<?> list, String property) {
...
+    public record JsonEntry(String first, Object last) {
+    }
+
+    /**
+     * Jekyll's "sort" filter for JsonObject (map-of-maps, e.g. data/authors.yaml).
+     * Returns a list of {@link JsonEntry} sorted by the named property on each value,
+     * so templates can use {data.first} (key) and {data.last} (value).
+     * Usage in Qute: {cdi:authors.sort('name')}
+     */
+    static List<JsonEntry> sort(JsonObject obj, String property) {
```
jtama, line 195: Ok, just to be sure.

If the given object is a map of json object, the result is a list and entries, sorted by the given property, if it exists, and if not, just a flatten list.

Is this correct ?

ia3andy (approved):
Looks good!

One minor suggestion: the javadoc additions add a lot of diff noise to this PR. The existing inline comments were concise and sufficient for a migration compatibility layer. Consider keeping them lean (or splitting the javadoc into a separate PR) to make functional changes easier to review.

holly-cummins:
I know what you mean about the Javadoc. I *almost* reverted all the javadoc noise, and then I decided that having the comments in a javadoc format probably was better overall, so I left it. I probably should have pulled it out to its own PR, though.

PR #1226, opened 2026-09-10
Converter: Ensure string literals have quotes when used as arguments

holly-cummins:
Another converter bug. We were generating some templates like `authors.sort(name)`. It needs to be `authors.sort('name')`.


jtama (approved):
Good

Issue #1222, opened 2026-09-09
Extract roq-resume and roq-linktree default layouts to reuse roq-base/default


and-y:
## Context

Raised in review on #1212: `roq-resume` and `roq-linktree` each define their own full `<html>...</html>` layout in `theme-layouts/*/default.html`, instead of extending `roq-base/default` the way the `roq-default` theme does (`layout: roq-base/default`).

This duplicates the base document shell (`<!DOCTYPE html>`, `<head>` boilerplate, `<html lang="...">` resolution, etc.) across three files, so any future change to that shell has to be applied in each theme separately.

## Proposal

Refactor `roq-theme/resume` and `roq-theme/linktree`'s `theme-layouts/*/default.html` to extend `roq-base/default` instead of redefining the full document, similar to how `roq-theme/default` does it.

## Out of scope for this issue

The theme-specific `<head>` content (favicon, SEO tags, bundle) and body classes/styling should stay theme-specific; only the shared document shell should move to `roq-base`.

and-y:
@ia3andy, looking at `roq-linktree`'s current `default.html` before refactoring it to extend `roq-base/default`: it applies `bg-slate-100 dark:bg-slate-900` to both `<html>` and `<body>` (the `<html>` one covers Safari's elastic-overscroll edges in dark mode, since the body already sets `overscroll-none`).

`roq-base/default`'s `<html>` tag has no class extension point today: only `<body>` supports a per-page class via `{#if page.data.body-class??}class="{page.data.body-class}"{/}`.

I'd like to add a symmetric `page.data.html-class` insert point to `roq-base/default`'s `<html>` tag, mirroring the existing `body-class` mechanism, so linktree can extend the shared shell without redefining it. Any objection before I go ahead with that?

PR #1231, opened 2026-09-11
Extract roq-resume and roq-linktree default layouts to reuse roq-base/default

and-y:
## Summary

- `roq-resume` and `roq-linktree`'s `theme-layouts/*/default.html` now extend `roq-base/default` (`layout: roq-base/default`) instead of redefining the full `<html>`/`<head>`/`<body>` document shell, matching the pattern already used by `roq-default`.
- `roq-base/default`'s `<html>` tag gained a symmetric `page.data.html-class` insert point (mirroring the existing `body-class` mechanism), needed because `roq-linktree` applies its overscroll-edge background classes to `<html>` itself, not just `<body>` (discussed on #1222, approved by @ia3andy).
- `roq-resume`'s `bodyClass` frontmatter key was renamed to `body-class` to match the key `roq-base`'s `<body>` tag actually reads.
- `roq-resume`'s remaining `head-meta` override was dropped too: it only duplicated what `roq-base`'s default `head-meta` block already provides (charset, viewport, seo, favicon, rss, bundle). Only the GA4 snippet and `<base href>` were theme-specific, so those moved into the additive `head` insert point, and the now-unused `partials/roq-resume/head.html` was removed.

Rendered output was diffed against the pre-refactor templates for every affected page (resume: `/`, `/404.html`; linktree: `/`, `/trees`, `/trees/test-links/`) — identical content, aside from insignificant formatting differences (attribute ordering/style) now coming from the shared base markup instead of each theme's own copy.

Refs #1222


ia3andy:
**Suggestion: `classNames` template extension + backward compat + documentation**

A couple of thoughts on the `body-class` / `html-class` approach:

1. **Backward compatibility on `bodyClass`**: The rename from `bodyClass` to `body-class` is a silent breaking change for existing users. We should support both keys during a transition period.

2. **`classNames` template extension**: Right now, FM class keys use replacement semantics (child overrides parent). For CSS classes, users typically expect additive behavior. We could add a `classNames` method to `RoqTemplateGlobal` that filters nulls/blanks and joins with spaces, e.g.:

   ```html
   <body class="{classNames(page.data('body-class'), page.data('bodyClass'), 'default')}">
   ```

   This solves both problems at once: backward compat for `bodyClass` and additive class composition for `html-class` / `body-class` across theme and page frontmatter.

3. **Documentation**: The new `body-class`, `html-class` keys and the `classNames` helper should be documented.

ia3andy:
That's a very nice refactor!! :)

and-y:
@ia3andy, `classNames` as a plain method on `RoqTemplateGlobal` doesn't compile: `@TemplateGlobal` only allows zero-argument methods, and `classNames` needs multiple.

The only Qute mechanism for a function like this, with no single receiver, is a namespace `@TemplateExtension`. I tried it on `RoqTemplateExtension` as `roq:classNames(...)` instead of a bare `classNames(...)`. There's precedent in the repo (`migration`'s `list:`/`str:` extensions), but it'd be a new namespace for `roq-frontmatter` itself. Any preference, or is `roq:` fine?

ia3andy:
Yeah let's introduce `roq:` 👌

and-y:
@ia3andy, one gap before finalizing: `page.data('body-class')` is already post-merge (`RoqFrontMatterStep3DataProcessor.mergeParents`, shallow `mergeIn`), so a page's value overwrites its layout's for the same key rather than combining with it. Tested it: layout `body-class: layout-default` + page `body-class: page-value` renders `"page-value"`, not `"layout-default page-value"`.

So `classNames` can only combine different keys, not a layout default with a page override of the same key. True cross-layer combination would need changing the FrontMatter merge itself, which I'm not touching as part of this refactor.

ia3andy:
Yeah that make sense, it was Claude suggesting that I thought why not but haven't thought it through.. anyway user can manually do whatever they want..

roq-frontmatter/runtime/src/main/resources/templates/theme-layouts/roq-base/default.html
```diff
@@ -1,7 +1,7 @@
 {@io.quarkiverse.roq.frontmatter.runtime.model.Page page}
 {@io.quarkiverse.roq.frontmatter.runtime.model.Site site}
 <!DOCTYPE html>
-<html lang="{page.data.getString('lang') ?: site.data.getString('lang') ?: global:htmlLocale}">
+<html lang="{page.data.getString('lang') ?: site.data.getString('lang') ?: global:htmlLocale}"{#let htmlClass=roq:classNames(page.data('html-class'), page.data('htmlClass'))} {#if htmlClass}class="{htmlClass}"{/if}{/let}>
```
ia3andy: we should be able to create an util to do that in one go?

and-y:
Added `roq:classAttr(...)`: wraps `classNames` and returns the full `class="..."` attribute (or `""`), so both tags collapse to one call each.

ia3andy:
Cool right? :)

and-y:
Yeah better now :)

ia3andy:
I found one issue: the deprecated keys (`htmlClass`/`bodyClass`) are combined additively with the new keys (`html-class`/`body-class`) instead of being used as fallbacks. This causes:

- Resume theme's `body-class: default` leaking a spurious `default` class when a page uses the old `bodyClass`
- `bodyClass` appearing on both `<body>` (via roq-base) and the wrapper `<div>` (via roq-default)
- Future removal of the deprecated keys becoming a breaking change instead of a silent no-op

Fix: use fallback semantics instead of additive:

```html
<html ...{roq:classAttr(page.data('html-class', page.data('htmlClass'))).raw}>
...
<body{roq:classAttr(page.data('body-class', page.data('bodyClass'))).raw}>
```

This way `html-class` wins if set, otherwise falls back to `htmlClass`. Clean migration path.

ia3andy (changes requested):
one small issue to fix 👆

and-y:
Rewrote this since your last look, squashed into one commit on current `main`.

**Why the previous approach went away.** Resolving the deprecated keys at render time, via three new `roq:` extensions (`classNames`, `classAttr`, `firstNonBlank`), ran into:

- Combining both keys leaked `roq-resume`'s own `body-class: default` onto any page still using `bodyClass`.
- Fallback instead of combining didn't help: a theme setting its own `body-class` default means the key is never absent, so the page's deprecated key stayed masked whatever the template did.
- `classAttr` emitted the `class="..."` string itself, quotes included, bypassing Qute's escaping: `body-class: x" onload="alert(1)` rendered as `<body class="x" onload="alert(1)">`.
- A single `null` expression passed to a varargs extension crashes Qute's varargs resolution, giving 500s on any page setting neither key.

Each fix was sound alone, but that is a lot of machinery for what #1222 asks.

**What it does now.** `RoqFrontMatterAssembleUtils.processTemplate` copies `htmlClass`/`bodyClass` to `html-class`/`body-class` per template, before the layout chain merges, warning with the file name. Templates read one key, so `roq-base/default` keeps the shape it already had:

```html
<body{#if page.data.body-class??} class="{page.data.body-class}"{/if}>
```

No extensions, no raw output, no manual escaping, nothing to crash on nulls; `RoqTemplateExtension` is untouched. Normalizing per template also fixes the masking: a page using `bodyClass` overrides its layout's default because by merge time it *is* a `body-class`, which no template-level fallback could do. `roq-default`'s `404`/`tag` layouts move to `body-class` so the theme stops warning about itself.

ia3andy (approved):
Thanks!!

PR #1244, opened 2026-09-16
feat(sitemap): add an HTML sitemap for visitors

### LINKED ISSUE #1213: Add ability to generate html sitemap
Roq can generate xml sitemaps. Sometimes it is also useful to have an html sitemap. The need for this is less clear (https://www.reddit.com/r/SEO/comments/11qi7ce/does_creating_a_html_sitemap_really_bring_any/ was the most recent discussion I can find), but it should be a straightforward extension of the xml-sitemap generation logic. 

The quarkus.io currently has a hand-edited html sitemap, and it would be nice to get rid of it.

and-y:
Adds an HTML sitemap alongside the XML one.

Fixes #1213

The plugin ships a second Qute fragment, `fm/sitemap.html`, included the
same way as the XML one:

```html
---
title: Sitemap
layout: page
---
{#include fm/sitemap.html /}
```

It renders loose pages first, then one `<section>` per visible collection.
Styling hooks: `.roq-sitemap`, `.roq-sitemap-pages`,
`.roq-sitemap-collection` and `data-collection="<id>"`. Empty groups are
dropped, and the page holding the include never links to itself.

It reuses the `sitemap:` FM flag, so `sitemap: false` excludes a page from
both outputs; there is no HTML-only exclusion.


ia3andy:
Should we also style it as part of the default theme?

ia3andy:
Also can you add it the blog?

and-y:
Adding to blog is a good idea + styled, will lock into it.

and-y:
Looked into both and implemented them:

- Styling: added `.roq-sitemap` / `.roq-sitemap-group` styles to the default theme, with bullet markers and a responsive column layout.
- Blog: added the sitemap at `/sitemap/` (`blog/content/sitemap.html`), linked from the nav menu.

blog/data/menu.yml
```diff
@@ -14,6 +14,9 @@ nav:
   - title: "Roqers"
     path: "/roqers/"
     icon: "fa fa-guitar"
+  - title: "Sitemap"
```
ia3andy: not needed?

and-y: My understanding was to add it. Are you asking to remove it?

ia3andy: Not as part of the menu

roq-plugin/sitemap/runtime/src/main/java/io/quarkiverse/roq/plugin/sitemap/runtime/RoqPluginSitemapTemplateExtension.java
```diff
@@ -25,4 +30,32 @@ public static boolean sitemap(Page page) {
...
+    /**
+     * The pages to list in the sitemap, excluding the one the sitemap itself is rendered from.<br>
+     * Example: "{#for p in site.sitemapPages(page)}".
+     */
+    public static List<NormalPage> sitemapPages(Site site, Page current) {
+        return site.pages().stream()
+                .filter(p -> p.source().isTargetHtml() && sitemap(p) && !p.id().equals(current.id()))
+                .toList();
+    }
+
```
ia3andy, line 42: `sitemapCollections()` creates a full filtered list via `sitemapDocuments(c)` just to check emptiness, then the template calls `collection.sitemapDocuments` again to iterate. You can avoid the throwaway list by using `anyMatch` to short-circuit:

```java
.filter(c -> !c.hidden() && !c.derived() && c.stream().anyMatch(RoqPluginSitemapTemplateExtension::sitemap))
```

ia3andy (changes requested):
Minor efficiency nit, otherwise LGTM.

roq-plugin/sitemap/runtime/src/main/java/io/quarkiverse/roq/plugin/sitemap/runtime/RoqPluginSitemapTemplateExtension.java
and-y: Yes good little optimization.

ia3andy (approved):
LGTM

PR #1246, opened 2026-09-17
docs: add offline usage instructions for the Roq CLI

### LINKED ISSUE #1217: Roq cli fails offline
If I try a build without network access (like on the Eurostar :) ), I get

```
holly@hcummins-mac quarkusio.github.io % roq build
[jbang] [ERROR] Could not download https://github.com/quarkiverse/quarkus-roq/blob/HEAD/jbang-catalog.json
[jbang] Run with --verbose or -x for more details. The --verbose or -x must be placed before the jbang command. I.e. jbang --verbose run [...]
```

I know jbang has an offline mode for dependencies, but I'm not sure that would help here, because it's the actual getting of the catalog that's failing.

matheusandre1:
Closes: #1217 

The error doesn't originate from the Roq CLI itself, but rather from the launcher generated by `jbang app install`. This launcher executes `jbang run roq@quarkiverse/quarkus-roq`, and jbang resolves this alias by fetching the catalog from GitHub. The catalog is cached for 12 hours (default `cache-evict` of PT12H). After that period, if there is no network connection, jbang attempts to download it again and fails, without falling back to the cached copy. Since this behavior is inherent to jbang, it cannot be fixed within the code of this repository; I have opened an issue for it [here](https://github.com/jbangdev/jbang/issues/2659) and will wait for a response before attempting a fix.


ia3andy:
@maxandersen is that a jbang issue? Jbang commands should work offline.

I think don't think adding instructions for something that should just be working is the right way forward..

maxandersen:
Just to be clear - this issue is s problem because GitHub is being down - so the requested data is not there. this is what --offline flag is for. 

That said we can probably consider simplifying it for catalogs you already downloaded...

